PR #2, opened 2016-08-30
set import path in .travis.yml; fix import-boss

lavalamp:


lavalamp:
This is going to fail travis, but I'm going to merge anyway-- protobuf imports an illegal package.

PR #3, opened 2016-08-30
Travis

lavalamp:


lavalamp:
The godoc badge will start working when we push the redirector.

PR #5, opened 2016-09-03
Channels deserve names, too

lavalamp:


thockin:
LGTM

thockin:
Travis failed.

lavalamp:
Failure is due to making an illegal import, pre existing problem.

thockin:
Does that mean we should merge this?  I'm LGTM, so go ahead if that is the right thing.

lavalamp:
Yeah, we should get these imports fixed though.

PR #6, opened 2016-09-06
Only add to tracker if it's non-local

lavalamp:
Import tracker was supposed to handle this case, but it seems local package was never plumbed through. Maybe lost in the protobuf addition?

### DIFF
--- a/namer/namer.go
+++ b/namer/namer.go
@@ -301,10 +301,10 @@ func (r *rawNamer) Name(t *types.Type) string {
 	if t.Name.Package != "" {
 		var name string
 		if r.tracker != nil {
-			r.tracker.AddType(t)
 			if t.Name.Package == r.pkg {
 				name = t.Name.Name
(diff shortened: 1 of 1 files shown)


lavalamp:
@smarterclayton I vaguely recall you saying something about this.

smarterclayton:
Wow that's so long ago I don't even know.  I think I ended up adding this all over the place in protobuf though.

lavalamp:
Yeah, the whole point was that the naming system should be doing this for you. But somehow the NewImportTracker method in the generator package doesn't let you specify your local package. Really that's where it should be fixed, but this is OK too.

thockin:
Is this worth re-considering?

lavalamp:
Yes, this should be fixed. I almost want to fork this whole package and fix everything :(

thockin:
rename it to Vampire because it has no reflection :)

PR #10, opened 2016-09-25
Add a defaulter-gen that coalesces individual defaulters

smarterclayton:
Creates top level SetObjectDefaults_ methods for specific types using
individual type defaulter functions (named `SetDefaults_TYPE`).

xref kubernetes/kubernetes#33352

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -1,0 +1,747 @@
+/*
+Copyright 2016 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generators
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"k8s.io/gengo/args"
+	"k8s.io/gengo/generator"
+	"k8s.io/gengo/namer"
+	"k8s.io/gengo/types"
+
+	"github.com/golang/glog"
+)
+
+// CustomArgs is used tby the go2idl framework to pass args specific to this
+// generator.
+type CustomArgs struct {
+	ExtraPeerDirs []string // Always consider these as last-ditch possibilities for conversions.
+}
+
+// This is the comment tag that carries parameters for defaulter generation.
+const tagName = "k8s:defaulter-gen"
+
+func extractTag(comments []string) []string {
+	return types.ExtractCommentTags("+", comments)[tagName]
+}
+
+func checkTag(comments []string, require ...string) bool {
+	values := types.ExtractCommentTags("+", comments)[tagName]
+	if len(require) == 0 {
+		return len(values) == 1 && values[0] == ""
+	}
+	return reflect.DeepEqual(values, require)
+}
+
+func defaultFnNamer() *namer.NameStrategy {
+	return &namer.NameStrategy{
+		Prefix: "SetDefaults_",
+		Join: func(pre string, in []string, post string) string {
+			return pre + strings.Join(in, "_") + post
+		},
+	}
+}
+
+func objectDefaultFnNamer() *namer.NameStrategy {
+	return &namer.NameStrategy{
+		Prefix: "SetObjectDefaults_",
+		Join: func(pre string, in []string, post string) string {
+			return pre + strings.Join(in, "_") + post
+		},
+	}
+}
+
+// NameSystems returns the name system used by the generators in this package.
+func NameSystems() namer.NameSystems {
+	return namer.NameSystems{
+		"public":          namer.NewPublicNamer(1),
+		"raw":             namer.NewRawNamer("", nil),
+		"defaultfn":       defaultFnNamer(),
+		"objectdefaultfn": objectDefaultFnNamer(),
+	}
+}
+
+// DefaultNameSystem returns the default name system for ordering the types to be
+// processed by the generators in this package.
+func DefaultNameSystem() string {
+	return "public"
+}
+
+// defaults holds the declared defaulting functions for a given type (all defaulting functions
+// are expected to be func(1))
+type defaults struct {
+	// object is the defaulter function for a top level type (typically one with TypeMeta) that
+	// invokes all child defaulters. May be nil if the object defaulter has not yet been generated.
+	object *types.Type
+	// base is a defaulter function defined for a type SetDefaults_Pod which does not invoke all
+	// child defaults - the base defaulter alone is insufficient to default a type
+	base *types.Type
+	// additional is zero or more defaulter functions of the form SetDefaults_Pod_XXXX that can be
+	// included in the Object defaulter.
+	additional []*types.Type
+}
+
+// All of the types in conversions map are of type "DeclarationOf" with
+// the underlying type being "Func".
+type defaulterFuncMap map[*types.Type]defaults
+
+// Returns all manually-defined defaulting functions in the package.
+func getManualDefaultingFunctions(context *generator.Context, pkg *types.Package, manualMap defaulterFuncMap) {
+	buffer := &bytes.Buffer{}
+	sw := generator.NewSnippetWriter(buffer, context, "$", "$")
+
+	for _, f := range pkg.Functions {
+		if f.Underlying == nil || f.Underlying.Kind != types.Func {
+			glog.Errorf("Malformed function: %#v", f)
+			continue
+		}
+		if f.Underlying.Signature == nil {
+			glog.Errorf("Function without signature: %#v", f)
+			continue
+		}
+		signature := f.Underlying.Signature
+		// Check whether the function is defaulting function.
+		// Note that all of them have signature:
+		// object: func SetObjectDefaults_inType(*inType)
+		// base: func SetDefaults_inType(*inType)
+		// additional: func SetDefaults_inType_Qualifier(*inType)
+		if signature.Receiver != nil {
+			continue
+		}
+		if len(signature.Parameters) != 1 {
+			continue
+		}
+		if len(signature.Results) != 0 {
+			continue
+		}
+		inType := signature.Parameters[0]
+		if inType.Kind != types.Pointer {
+			continue
+		}
+		// Check if this is the primary defaulter.
+		args := defaultingArgsFromType(inType.Elem)
+		sw.Do("$.inType|defaultfn$", args)
+		switch {
+		case f.Name.Name == buffer.String():
+			key := inType.Elem
+			// We might scan the same package twice, and that's OK.
+			v, ok := manualMap[key]
+			if ok && v.base != nil && v.base.Name.Package != pkg.Path {
+				panic(fmt.Sprintf("duplicate static defaulter defined: %#v", key))
+			}
+			v.base = f
+			manualMap[key] = v
+			glog.V(6).Infof("found base defaulter function for %s from %s", key.Name, f.Name)
+		// Is one of the additional defaulters - a top level defaulter on a type that is
+		// also invoked.
+		case strings.HasPrefix(f.Name.Name, buffer.String()+"_"):
+			key := inType.Elem
+			v, ok := manualMap[key]
+			if ok {
+				exists := false
+				for _, existing := range v.additional {
+					if existing.Name == f.Name {
+						exists = true
+						break
+					}
+				}
+				if exists {
+					continue
+				}
+			}
+			v.additional = append(v.additional, f)
+			manualMap[key] = v
+			glog.V(6).Infof("found additional defaulter function for %s from %s", key.Name, f.Name)
+		}
+		buffer.Reset()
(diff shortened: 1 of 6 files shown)


smarterclayton:
Ready for review

lavalamp:
@mbohlool any chance you can take this review?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,721 @@
...
+)
+
+// CustomArgs is used tby the go2idl framework to pass args specific to this
+// generator.
+type CustomArgs struct {
+	ExtraPeerDirs []string // Always consider these as last-ditch possibilities for conversions.
+}
+
+// This is the comment tag that carries parameters for conversion generation.
+const tagName = "k8s:defaulter-gen"
```
thockin: we should probably parameterize tag names or nix the k8s: prefix in favor of ... ?

smarterclayton: Parameterizing would be easiest.

smarterclayton:
Examples are https://github.com/kubernetes/kubernetes/pull/33352/commits/b08faa9da86cf7be4c936f37b6c5bcb3f8c32d4a

smarterclayton:
Practically this is a baby step towards true defaulter generation.  This is enough to get us top level defaulters per type as a separate code path without a full redesign of defaulting.

smarterclayton:
@wojtek-t as well if he has time

smarterclayton:
@wojtek-t this is the prereq for the unsafe conversion stuff

smarterclayton:
For future reference `map[reflect.Type]func(interface{})` is about 50ns faster than calling 5 functions that have type switches (`switch foo.(type) { case *Type1: CallFunc(t) }`).  I think given that I'm going to change the "fast conversion" registration in scheme to be a type pair map:

```
type typePair struct {
  from, to reflect.Type
}
type Convertor struct {
  conversions map[typePair]func(from, to interface{}, scope) error
}

convertor.Register(&Type{}, func(a, b interface{}, s scope) error { return Convert_X_to_Y(a.(*TypeA), b.(*TypeB), scope) })
```

Which is more composable than our current struct.

smarterclayton:
Another option here is to add `.SetDefaults()` onto each object as a method (instead of having to register in a scheme).  We've gradually relaxed our requirements around objects - this would be a natural next step (although we'd have to pick a name that wouldn't conflict with any field we added in the future).  That would make defaulting a simple interface check and remove the need for wrappers.  I can imagine that a lot of generic code in the versioned types would prefer an accessor.

lavalamp:
I don't like that option because it means that code very likely won't even
build if the generated files aren't present.

smarterclayton:
Isn't that a good thing?  "You aren't getting defaulting, fix that" is different from "We were going to default you, but nothing happened, so you got no defaults".  The API server is broken without defaulting.  Certainly nothing in client-go should be depending on Defaults().

lavalamp:
But:
* Not all users want defaulting (clients). Compiling in code they don't
want adds unwanted bulk and more importantly, dependencies.
* Editing tools / IDEs don't need defaulting and should work correctly
whether or not the files have been generated.

I think it's OK if apiserver doesn't build without these files, but it'd be
nice if that compile error came from one place rather than being
distributed lots of places.

smarterclayton:
Is *all* of the complexity of having it be outside worth avoiding the use of either a go-build tag or just not using it?

Defaulters require zero dependencies at the current time.

smarterclayton:
Note that I'm not in total disagreement - but I suspect that 99% of developers would just generate a SetDefaults() method onto the versioned types and call it a day.  Are we really really clever?  Or not as clever as we think?  :)

lavalamp:
I guess the thing is, for defaults specifically, I don't want the
SetDefaults method to even be available from the client library, because we
don't want clients to call it. The server is the source of defaults.
Therefore we need to have some convenient way of not including it in the
client library.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+	"github.com/golang/glog"
+)
+
+// CustomArgs is used tby the go2idl framework to pass args specific to this
+// generator.
+type CustomArgs struct {
+	ExtraPeerDirs []string // Always consider these as last-ditch possibilities for conversions.
+}
+
+// This is the comment tag that carries parameters for conversion generation.
```
wojtek-t: s/conversion/defaulting/

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+	values := types.ExtractCommentTags("+", comments)[tagName]
+	if len(require) == 0 {
+		return len(values) == 1 && values[0] == ""
+	}
+	return reflect.DeepEqual(values, require)
+}
+
+// TODO: This is created only to reduce number of changes in a single PR.
+// Remove it and use PublicNamer instead.
+func conversionNamer() *namer.NameStrategy {
```
wojtek-t: Should we name it "conversionNamer" ?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+// DefaultNameSystem returns the default name system for ordering the types to be
+// processed by the generators in this package.
+func DefaultNameSystem() string {
+	return "public"
+}
+
+type defaults struct {
+	object     *types.Type
+	base       *types.Type
+	additional []*types.Type
```
wojtek-t: Can you please add comments what are "object", "base" and "additional"?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+			continue
+		}
+		if f.Underlying.Signature == nil {
+			glog.Errorf("Function without signature: %#v", f)
+			continue
+		}
+		signature := f.Underlying.Signature
+		// Check whether the function is conversion function.
+		// Note that all of them have signature:
+		// func Convert_inType_To_outType(inType, outType, conversion.Scope) error
```
wojtek-t: The signature is different for defaulting functions.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+		if f.Underlying == nil || f.Underlying.Kind != types.Func {
+			glog.Errorf("Malformed function: %#v", f)
+			continue
+		}
+		if f.Underlying.Signature == nil {
+			glog.Errorf("Function without signature: %#v", f)
+			continue
+		}
+		signature := f.Underlying.Signature
+		// Check whether the function is conversion function.
```
wojtek-t: s/conversion/defaulting/

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+			key := inType.Elem
+			// We might scan the same package twice, and that's OK.
+			v, ok := manualMap[key]
+			if ok && v.base != nil && v.base.Name.Package != pkg.Path {
+				panic(fmt.Sprintf("duplicate static defaulter defined: %#v", key))
+			}
+			v.base = f
+			manualMap[key] = v
+			glog.V(6).Infof("found base defaulter function for %s from %s", key.Name, f.Name)
+		// Is one of the additional defaulters
```
wojtek-t: What is "additional defaulter"? Can you please clarify?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+			}
+			// opt-out
+			if checkTag(t.SecondClosestCommentLines, "false") {
+				return false
+			}
+			// opt-in
+			if checkTag(t.SecondClosestCommentLines, "true") {
+				return true
+			}
+			// type contains a field with a type name matching the pkg comment tag
```
wojtek-t: What's the usecase for that? [I'm not saying it's unneeded - just want to understand it]

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+		packages = append(packages,
+			&generator.DefaultPackage{
+				PackageName: filepath.Base(pkg.Path),
+				PackagePath: pkg.Path,
+				HeaderText:  header,
+				GeneratorFunc: func(c *generator.Context) (generators []generator.Generator) {
+					generators = []generator.Generator{}
+					generators = append(
+						generators, NewGenDefaulter(arguments.OutputFileBaseName, pkg.Path, existingDefaulters, newDefaulters, peerPkgs))
+					return generators
```
wojtek-t: nit:
```
return []generator.Generator{NewGenDefaulter(argument...)}
```

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+}
+
+func (g *genDefaulter) generateDefaulter(inType *types.Type, callTree *callNode, sw *generator.SnippetWriter) {
+	sw.Do("func $.inType|objectdefaultfn$(in *$.inType|raw$) {\n", defaultingArgsFromType(inType))
+	callTree.WriteMethod("in", 0, nil, sw)
+	sw.Do("}\n\n", nil)
+}
+
+type callNode struct {
+	field string
```
wojtek-t: Can you please add comments to the fields (it's not clear to me what values they are storing).

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+	key   bool
+	index bool
+	elem  bool
+
+	call     []*types.Type
+	children []callNode
```
wojtek-t: IIUC, children contains "recursive calls" of defaulting function. So what are "call"?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,710 @@
...
+				},
+				FilterFunc: func(c *generator.Context, t *types.Type) bool {
+					return t.Name.Package == pkg.Path
+				},
+			})
+	}
+	return packages
+}
+
+func buildCallTreeForType(t *types.Type, depth int, existingDefaulters, newDefaulters defaulterFuncMap) *callNode {
```
wojtek-t: Why do you need "depth"?

examples/defaulter-gen/generators/defaulter.go
smarterclayton: Conversion assumes all objects need conversion in a package except those that opt out (reasonable - we match everything to everything).  OpenAPI is similar.  Defaulting, on the other hand, is used almost exclusively on top level objects, and our defaulters today do not "cover" (to properly default you have to invoke all nested defaulters at the right spot).

This code generates top level covering defaulters for everything that needs them.  So PodSpec doesn't really need a covering defaulter - but Pod does.  So instead of the blanket package inclusion + selective exclusion, the defaulter supports selective inclusion/exclusion per type, and then blanket selectors based on whether you have a field.  In Kube, conveniently, every type we want a top level defaulter for has TypeMeta.  Rather than hardcode type meta, I allow it to be defined as a field.  So all Lists + Objects + quasi-objects (RangeAllocations) etc get object defaulters.

examples/defaulter-gen/generators/defaulter.go
smarterclayton: Depth here is just to know whether we are at the root.

smarterclayton: I'll change it to a bool.

examples/defaulter-gen/generators/defaulter.go
smarterclayton: calls on that particular field (adding an example and better godoc now)

smarterclayton:
Added much, much more Godoc.  PTAL

wojtek-t:
@smarterclayton - thanks; will take another look tomorrow

smarterclayton:
As discussed on sig-api-machinery no change to defaulter behavior for now - we're still pretty comfortable keeping things outside the core objects and Dan's wish to not have defaulters available for some clients is reasonable.

examples/defaulter-gen/generators/defaulter.go
smarterclayton: Also this supports multiple field coverage by specifying multiple times.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -0,0 +1,744 @@
...
+		// we can halt recursion
+		if checkTag(defaults.base.CommentLines, "covers") {
+			glog.V(6).Infof("the defaulter %s indicates it covers all sub generators", t.Name)
+			return parent
+		}
+	}
+
+	for _, fn := range defaults.additional {
+		parent.call = append(parent.call, fn)
+	}
```
wojtek-t: nit: why not just:
```
parent.call = append(parent.call, defaults.additional...)
```
?

wojtek-t:
@smarterclayton - two minor comments, other than that LGTM

>
As discussed on sig-api-machinery no change to defaulter behavior for now (method on object vs method separate, as implemented here) - we're still pretty comfortable keeping things outside the core objects and Dan's wish to not have defaulters available for some clients is reasonable.

Good to know. Though I still think that we should merge this PR, even if we are not planning to use it in kubernetes for now. WDYT?

smarterclayton:
Sorry, we are planning to use this - that comment was about not changing
how this PR works to generate `func (*Pod) SetDefaults()` instead of `func
SetObjectDefaults(*pod)`

examples/defaulter-gen/generators/defaulter.go
smarterclayton: Yeah

smarterclayton:
Updated with comments, tested this against my master PR - merging

wojtek-t:
LGTM

PR #12, opened 2016-10-19
Extract function comments

smarterclayton:
Allows generators like defaulters and conversions to use information
about generated conversions.

@wojtek-t extracted from the unsafe conversions work

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -453,6 +453,7 @@ func (b *Builder) findTypesIn(pkgPath string, u *types.Universe) error {
 		if ok {
 			t := b.walkType(*u, nil, tn.Type())
 			c1 := b.priorCommentLines(obj.Pos(), 1)
+			// c1.Text() is safe if c1 is nil
 			t.CommentLines = splitLines(c1.Text())
 			if c1 == nil {
 				t.SecondClosestCommentLines = splitLines(b.priorCommentLines(obj.Pos(), 2).Text())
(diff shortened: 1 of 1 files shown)


parser/parse.go
```diff
@@ -463,7 +463,14 @@ func (b *Builder) findTypesIn(pkgPath string, u *types.Universe) error {
 		tf, ok := obj.(*tc.Func)
 		// We only care about functions, not concrete/abstract methods.
 		if ok && tf.Type() != nil && tf.Type().(*tc.Signature).Recv() == nil {
-			b.addFunction(*u, nil, tf)
+			t := b.addFunction(*u, nil, tf)
+			c1 := b.priorCommentLines(obj.Pos(), 1)
+			t.CommentLines = splitLines(c1.Text())
+			if c1 == nil {
```
thockin: is it kosher to call c1.Text() on the previous line if c1 is nil?  If you have a convincing argument, comment it.  else fix.

smarterclayton: Totally cribbed from 5 lines above, answer is yes, I can add comment in both places.

smarterclayton:
Updated

wojtek-t:
lgtm

thockin:
LGTM

wojtek-t:
This one is needed for https://github.com/kubernetes/kubernetes/pull/34784 so I'm merging this.

PR #15, opened 2016-10-26
Defaulter: emit no-op files

thockin:
The tag said to process a directory.  If we don't emit the result file, the
kubernetes Make sysyetm will detect that this file is missing and repeatedly
try to generate it on every build.

Once this goes in, I will re-vendor it into the kube repo.

@wojtek-t


wojtek-t:
LGTM - thanks TIm!

thockin:
I'm looking at Clayton for a final look.  This was already done in deepcopy

smarterclayton:
LGTM

PR #16, opened 2016-10-27
Protobuf: allow dashes in package names

ncdc:
Convert - to _ for protobuf package names to allow protobuf code generation
support for go packages that have - in their names.

xref https://github.com/kubernetes/kubernetes/pull/33412

cc @smarterclayton


smarterclayton:
As merged upstream.

ncdc:
Heh it hasn't merged in kube yet

smarterclayton:
LGTM

PR #18, opened 2016-11-05
Export unregistered deep copy functions, use custom DeepCopy impls

liggitt:
Makes use of custom DeepCopy functions if defined (so we can define DeepCopy for types with unexported fields, and not switch to shallow copy in the middle of deep copy)

Adds function that returns generated deepcopy functions if they are not registered (see use at https://github.com/kubernetes/kubernetes/pull/35728/commits/cd69db639ae098e96b371c7aa774679dc90a3acb, sample generated diff at https://github.com/kubernetes/kubernetes/pull/35728/commits/bfb3a7eb007476283870d7fd1f2e2b0d2df71927)

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/gengo/18)

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -546,7 +546,14 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 	if len(t.Members) == 0 {
 		// at least do something with in/out to avoid "declared and not used" errors
 		sw.Do("_ = in\n_ = out\n", nil)
+		return
 	}
+
+	if hasDeepCopyMethod(t) {
+		sw.Do("*out = in.DeepCopy()\n", nil)
+		return
+	}
+
 	for _, m := range t.Members {
 		t := m.Type
 		if t.Kind == types.Alias {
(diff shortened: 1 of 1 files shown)


liggitt:
cc @smarterclayton

liggitt:
@smarterclayton @wojtek-t PTAL

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -538,7 +546,15 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 	if len(t.Members) == 0 {
 		// at least do something with in/out to avoid "declared and not used" errors
 		sw.Do("_ = in\n_ = out\n", nil)
+		return
+	}
+
+	if hasDeepCopyMethod(t) {
+		sw.Do("outCopy := in.DeepCopy()\n", nil)
+		sw.Do("*out = outCopy\n", nil)
```
lavalamp: Why not `sw.Do("*out := in.DeepCopy()\n", nil)`?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -579,11 +595,17 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
...
-			sw.Do("}\n", nil)
+			if hasDeepCopyMethod(t) {
+				sw.Do("} else {\n", nil)
+				sw.Do("out.$.name$ = in.$.name$.DeepCopy()\n", args)
+				sw.Do("}\n", nil)
+			} else {
+				sw.Do("} else if newVal, err := c.DeepCopy(&in.$.name$); err != nil {\n", args)
+				sw.Do("return err\n", nil)
+				sw.Do("} else {\n", nil)
+				sw.Do("out.$.name$ = *newVal.(*$.type|raw$)\n", args)
```
lavalamp: should this line ever actually get executed?

lavalamp:
Seeing a diff of this applied to the main repo would make it easier to lgtm :)

liggitt:
diff with this applied: https://github.com/kubernetes/kubernetes/pull/35728/commits/be7606c8704ab86f7571a91b1927ebde941be495

examples/deepcopy-gen/generators/deepcopy.go
liggitt: I don't know? this is just preserving the existing code in the case where a type doesn't define a custom `DeepCopy()` method

examples/deepcopy-gen/generators/deepcopy.go
liggitt: updated

liggitt:
comments addressed

resulting diff in generated copies at https://github.com/kubernetes/kubernetes/pull/35728/commits/bfb3a7eb007476283870d7fd1f2e2b0d2df71927

usage of generated `GetGeneratedDeepCopyFuncs` functions at https://github.com/kubernetes/kubernetes/pull/35728/commits/cd69db639ae098e96b371c7aa774679dc90a3acb

lavalamp:
LGTM

PR #21, opened 2016-12-30
Accumulated cleanups, logging, and a test!

thockin:
As "leisure time" I wanted to see about applying gengo deepcopy to non-kube projects.  Not surprisingly, it didn't work in obvious ways, so I set out to fix it.  My list got long quickly, so here is part 1.

This represents some better logging to help me figure out what was not working, and I added a test to check "golden" output.  It's rudimentary for now, but it can expand.

PR #22, opened 2016-12-30
Optimize deepcopy for primitives

thockin:
As part of #21 (upon which this builds, only the last 2 commits are this PR proper), I realized deepcopy was being dumb.

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -131,6 +131,9 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 
 	boundingDirs := []string{}
 	if customArgs, ok := arguments.CustomArgs.(*CustomArgs); ok {
+		if customArgs.BoundingDirs == nil {
+			customArgs.BoundingDirs = context.Inputs
+		}
 		for i := range customArgs.BoundingDirs {
 			// Strip any trailing slashes - they are not exactly "correct" but
 			// this is friendlier.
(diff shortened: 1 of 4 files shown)


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -546,66 +549,85 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
...
 
+	// Simple copy covers a lot of cases.
+	sw.Do("*out = *in\n", nil)
+
+	// Now fix-up fields as needed.
 	for _, m := range t.Members {
 		t := m.Type
+		hasMethod := false
+		if hasDeepCopyMethod(t) {
+			hasMethod = true
```
wojtek-t: Why not just:
```
hasMethod := hasDeepCopyMethod(t)
```

?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -546,66 +549,85 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
...
 		case types.Map, types.Slice, types.Pointer:
-			sw.Do("if in.$.name$ != nil {\n", args)
-			sw.Do("in, out := &in.$.name$, &out.$.name$\n", args)
-			g.generateFor(t, sw)
-			sw.Do("} else {\n", nil)
-			sw.Do("out.$.name$ = nil\n", args)
-			sw.Do("}\n", nil)
+			if hasMethod {
+				sw.Do("if in.$.name$ != nil {\n", args)
+				sw.Do("    out.$.name$ = in.$.name$.DeepCopy()\n", args)
```
wojtek-t: We explicitly stopped adding spaces at the beginning, because we are calling gofmt at the end anyway. I think you should also remove them here and below - this is painful to maintain them.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -546,66 +549,85 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
...
 		}
 		args := generator.Args{
 			"type": t,
+			"kind": t.Kind,
 			"name": m.Name,
 		}
 		switch t.Kind {
-		case types.Builtin:
-			sw.Do("out.$.name$ = in.$.name$\n", args)
+		case types.Builtin, types.Array:
```
wojtek-t: I might be missing something about Go, but is adding Array here really a good idea? What if underlying type of an arrays is something complex that requires proper deep-copy?

wojtek-t:
Do we have merge-bot in this repo? Or should i merge it manually?

thockin:
I don't think we have merge bot

PR #23, opened 2016-12-30
Use strong type for pkg-paths when parsing

thockin:
This makes the code easier to comprehend.  This commit also sanitizes how
import paths are canonicalized, which in turn allows `./path` package names to
work.

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -46,7 +46,7 @@ type Builder struct {
 	absPaths map[string]string
 
 	// Set by makePackage(), used by importer() and friends.
-	pkgs map[string]*tc.Package
+	typeCheckedPackages map[string]*tc.Package
 
 	// Map of package path to whether the user requested it or it was from
 	// an import.
@@ -88,6 +88,7 @@ func New() *Builder {
 	return &Builder{
 		context:               &c,
 		buildInfo:             map[string]*build.Package{},
+		typeCheckedPackages:   map[string]*tc.Package{},
 		fset:                  token.NewFileSet(),
 		parsed:                map[string][]parsedFile{},
 		absPaths:              map[string]string{},
@@ -318,14 +319,14 @@ type importAdapter struct {
 }
 
 func (a importAdapter) Import(path string) (*tc.Package, error) {
-	return a.b.importer(a.b.pkgs, path)
+	return a.b.importer(a.b.typeCheckedPackages, path)
 }
 
 // typeCheckPackage will attempt to return the package even if there are some
 // errors, so you may check whether the package is nil or not even if you get
 // an error.
 func (b *Builder) typeCheckPackage(id string) (*tc.Package, error) {
-	if pkg, ok := b.pkgs[id]; ok {
+	if pkg, ok := b.typeCheckedPackages[id]; ok {
 		if pkg != nil {
 			return pkg, nil
 		}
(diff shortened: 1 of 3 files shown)


thockin:
I've done a deeper overhaul, which will have to be broken into smaller pieces.  This is what happens when I try to understand things :)

thockin:
OK, broke out all the things that were O(easy).  I'm willing to do more if need be, but the rest will be much harder, I think (or will not make the most egregious commit any better).

parser/parse.go
```diff
@@ -105,84 +114,86 @@ func (b *Builder) AddBuildTags(tags ...string) {
...
-func (b *Builder) buildPackage(pkgPath string) (*build.Package, error) {
-	// This is a bit of a hack.  The srcDir argument to Import() should
-	// properly be the dir of the file which depends on the package to be
-	// imported, so that vendoring can work properly.  We assume that there is
-	// only one level of vendoring, and that the CWD is inside the GOPATH, so
-	// this should be safe.
-	cwd, err := os.Getwd()
-	if err != nil {
-		return nil, fmt.Errorf("unable to get current directory: %v", err)
+func (b *Builder) importBuildPackage(dir string) (*build.Package, error) {
```
deads2k: I found this coming the other direction because it is improperly creating imports that look like this `v1 "k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/apis/meta/v1"` .  and that behavior appears to be intentional based on https://golang.org/src/go/build/build.go#L570

parser/parse.go
```diff
@@ -239,77 +245,100 @@ func (b *Builder) AddDirRecursive(dir string) error {
...
+	glog.V(5).Infof("addDir %s", dir)
+	buildPkg, err := b.importBuildPackage(dir)
 	if err != nil {
 		return err
 	}
-	// Check in case this package was added (maybe dir was not canonical)
-	if wasRequested, wasAdded := b.userRequested[dir]; wasAdded {
-		if !userRequested || userRequested == wasRequested {
-			return nil
+	pkgPath := importPathString(buildPkg.ImportPath)
```
deads2k: Yeah, this it here.  The `ImportPath` is *sometimes*, but not always the same as package name.  It seems to be intentionally not a valid package path based on construction here https://golang.org/src/go/build/build.go#L570 .  Still thinking about how to detect it.

thockin:
What problem is this causing?  We canonicalize import paths now, so the
entire gengo parser will convert arbitrary user-provided names "
k8s.io/apimachinery/pkg/apis/meta/v1" into "
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/apis/meta/v1" when storing
into maps and things like that.  Is that getting exposed somewhere?

deads2k:
> Is that getting exposed somewhere?

It's getting used as the `pkgPath` for import statements, which is failing when go tries to compile them.

thockin:
Interesting - that doesn't currently trigger in the main kube tree, I
guess?  Maybe we need to StripPrefix `k8s.io/kubernetes/vendor/` when we
canonicalize, though I am not sure we know that prefix, and maybe we need
to handle nested vendor trees?

maybe we need to strip the common prefix + `vendor` when we emit import
statements?

deads2k:
> maybe we need to strip the common prefix + `vendor` when we emit import
> statements?

Do we know the common prefix?

deads2k:
Created https://github.com/kubernetes/kubernetes/pull/39643 to demonstrate the failure.

thockin:
Commented on 39644 - it's similar to your hack, but only done at emit-time,
and a bit stricter checking.

PR #27, opened 2017-01-03
make import-boss attempt to look in the vendor folder

deads2k:
As we split out the genericapiserver, we need to enforce import requirements.  `staging/k8s.io/genericapiserver` is used as a vendored folder when building kube, so the `.import_restrictions` file exists in the vendor folder.

I know some effort went into removing kube references from this code, so I'm looking for recommendations to make this code generic.  I'm not very familiar with this repo.

I stumbled into this problem while working on https://github.com/kubernetes/kubernetes/pull/39384 .  I can commit that pull and continue for a while without `import-boss`, but I'd really like to have it.

### DIFF
--- a/examples/import-boss/generators/import_restrict.go
+++ b/examples/import-boss/generators/import_restrict.go
@@ -192,8 +192,16 @@ func (importRuleFile) VerifyFile(f *generator.File, path string) error {
 	}
 
 	if rules == nil {
-		// No restrictions on this directory.
-		return nil
+		// No restrictions on this directory, double check under vendor.  Make an assumption about running in kube
+		vendorPath := strings.Replace(path, "/go/src/k8s.io/", "/go/src/k8s.io/kubernetes/vendor/k8s.io/", -1)
+		rules, actualPath, err = recursiveRead(vendorPath)
+		if err != nil {
+			return fmt.Errorf("error finding rules file: %v", err)
+		}
+
+		if rules == nil {
+			return nil
+		}
 	}
 
 	for _, r := range rules.Rules {


deads2k:
@thockin @wojtek-t @lavalamp you guys look to be working in this repo.

examples/import-boss/generators/import_restrict.go
```diff
@@ -192,8 +192,16 @@ func (importRuleFile) VerifyFile(f *generator.File, path string) error {
 	}
 
 	if rules == nil {
-		// No restrictions on this directory.
-		return nil
+		// No restrictions on this directory, double check under vendor.  Make an assumption about running in kube
+		vendorPath := strings.Replace(path, "/go/src/k8s.io/", "/go/src/k8s.io/kubernetes/vendor/k8s.io/", -1)
```
thockin: I have never looked at the import boss but this is certainly wrong.  I'm not sure I understand why you want to look for rules in vendor?

lavalamp: Yeah, I don't understand why you want to do this, either.

deads2k: I can't seem to find a place that indicates the directory that the supporting files were found in, do the path always ends up k8s.io/kubernetes without vendor in it.

As I noted in the description, I know this isn't right, but this is the spot I need to know where to look for the .Import-path and I'm looking for a recommendation of what to use to find it.

lavalamp: I guess I don't understand what the problem is that this is fixing?

deads2k: Oh, the Import-path file here: https://github.com/kubernetes/kubernetes/pull/39384 isn't enforced because when I add it to the import-boss command, it looks in the wrong folder (not vendor) for the import-restriction file.  This forces it to check again in the correct directory, but isn't a reasonable solution.

deads2k:
I think we'll still want to be able to support import-boss in staging, but for now the simple golist verify script is working.

PR #28, opened 2017-01-04
Deepcopy tests for slices and manual DeepCopy methods

thockin:
Amazing the things you find when you test.

### DIFF
--- a/examples/deepcopy-gen/test/wholepkg/deepcopy_generated.go
+++ b/examples/deepcopy-gen/test/wholepkg/deepcopy_generated.go
@@ -29,8 +29,10 @@ import (
 func GetGeneratedDeepCopyFuncs() []conversion.GeneratedDeepCopyFunc {
 	return []conversion.GeneratedDeepCopyFunc{
 		{Fn: DeepCopy_wholepkg_ManualStruct, InType: reflect.TypeOf(&ManualStruct{})},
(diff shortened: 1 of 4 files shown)


wojtek-t:
One minor question - feel free to self-lgtm it.

thockin:
What was the question?  I think you forgot to ask it :)

wojtek-t:
Sorry - I was on vacation. You shouldn't have wait for me with this PR. And now I clicked 'Comment' finally.

examples/deepcopy-gen/test/wholepkg/deepcopy_generated.go
```diff
@@ -54,6 +60,15 @@ func DeepCopy_wholepkg_ManualStruct(in interface{}, out interface{}, c *conversi
 	}
 }
 
+func DeepCopy_wholepkg_ManualStruct_Alias(in interface{}, out interface{}, c *conversion.Cloner) error {
+	{
```
thockin: good question.  It's not in this PR, but I'll look for it.

thockin:
I'm gonna merge this and hunt that issue in a new PR

examples/deepcopy-gen/test/wholepkg/deepcopy_generated.go
thockin: This is needed to do the `in := in.(*cast)` stuff.  without the extra nesting, it doesn't compile.  It has been there for a long time.

Issue #30, opened 2017-01-09
gengo change of behavior with generated import statements


jpeeler:
When using gengo in service-catalog, I found that the import statements aren't generated in the same way any more. The change of behavior started after the commit of 7ca94dd123f0f4965aac2e59ef249d8eaef14bcc, which was on Jan 3. I filed a bug https://github.com/kubernetes-incubator/service-catalog/issues/223 showing the difference in behavior.

thockin:
Replied on your bug.

This is not a bug in gengo, I think.  Having it generate paths with `vendor` in them seems wrong.  The resulting change looks like you should need to add `k8s.io/kubernetes` to `--bounding-dirs` on deepcopy

jpeeler:
The committed code above appears to have fixed the problem.

PR #29, opened 2017-01-05
update apimachinery package to match current state

deads2k:
https://github.com/kubernetes/kubernetes/pull/39475 moves the apimachinery to a logically separate repo under staging.  This updates the packages to match what will be coming.

@lavalamp

PR #31, opened 2017-01-10
handle vendoring for canonicalization

deads2k:
The `.ImportPath` from `build` produces a file reference, not the canonical package import path.  This results in package imports like this: `pkgfoo "k8s.io/kubernetes/vendor/k8s.io/fake-dep/pkgfoo"`.

I attempted to handle this by rewriting the import just prior to serialization, but that doesn't handle cases where the package is used to inspect functions.  Because it doesn't rewrite the package path when its discovered, the package is wrong for any comparison against that package. In particular, this check here in conversion fails https://github.com/kubernetes/kubernetes/blob/master/cmd/libs/go2idl/conversion-gen/generators/conversion.go#L144, because it compares `k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/conversion.Scope` against `k8s.io/apimachinery/pkg/conversion.Scope`.

This patch will mis-identify a package that is actually called `vendor`, but it does correctly handle nesting at any level.  It does
 1.  `k8s.io/kubernetes/vendor/github.com/foo/bar` -> ` github.com/foo/bar`
 2. `k8s.io/contrib/some/vendor/github.com/foo/bar` -> ` github.com/foo/bar`
 3. `k8s.io/contrib/vendor/github.com/foo/bar` -> `github.com/foo/bar`

I think this is better than our current state today, we have no "vendor" package anywhere dep'ed in today and someone would have to be kind of crazy to make it.

@thockin @lavalamp

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -136,14 +136,15 @@ func (b *Builder) importBuildPackage(dir string) (*build.Package, error) {
 	// Remember it under the user-provided name.
 	glog.V(5).Infof("saving buildPackage %s", dir)
 	b.buildPackages[dir] = buildPkg
-	if dir != buildPkg.ImportPath {
+	canonicalPackage := canonicalizeImportPath(buildPkg.ImportPath)
+	if dir != string(canonicalPackage) {
 		// Since `dir` is not the canonical name, see if we knew it under another name.
-		if buildPkg, ok := b.buildPackages[buildPkg.ImportPath]; ok {
+		if buildPkg, ok := b.buildPackages[string(canonicalPackage)]; ok {
 			return buildPkg, nil
 		}
 		// Must be new, save it under the canonical name, too.
-		glog.V(5).Infof("saving buildPackage %s", buildPkg.ImportPath)
-		b.buildPackages[buildPkg.ImportPath] = buildPkg
+		glog.V(5).Infof("saving buildPackage %s", canonicalPackage)
+		b.buildPackages[string(canonicalPackage)] = buildPkg
 	}
 
 	return buildPkg, nil
@@ -223,7 +224,7 @@ func (b *Builder) AddDirRecursive(dir string) error {
 			rel := strings.TrimPrefix(path, prefix)
 			if rel != "" {
 				// Make a pkg path.
-				pkg := filepath.Join(b.buildPackages[dir].ImportPath, rel)
+				pkg := filepath.Join(string(canonicalizeImportPath(b.buildPackages[dir].ImportPath)), rel)
 
 				// Add it.
 				if _, err := b.importPackage(pkg, true); err != nil {
@@ -250,7 +251,7 @@ func (b *Builder) AddDirTo(dir string, u *types.Universe) error {
 	if _, err := b.importPackage(dir, true); err != nil {
 		return err
 	}
-	return b.findTypesIn(importPathString(b.buildPackages[dir].ImportPath), u)
+	return b.findTypesIn(canonicalizeImportPath(b.buildPackages[dir].ImportPath), u)
 }
 
 // The implementation of AddDir. A flag indicates whether this directory was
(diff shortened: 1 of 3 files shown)


deads2k:
Daniel pointed out via slack that "vendor" isn't a legal package name, so this doesn't have that gap.

parser/parse.go
```diff
@@ -749,3 +753,14 @@ func (b *Builder) addVariable(u types.Universe, useName *types.Name, in *tc.Var)
 	out.Underlying = b.walkType(u, nil, in.Type())
 	return out
 }
+
+// importPathToFullPackage takes an import path and returns the actual package
+// it doesn't work for any folder actually called `vendor` (we don't dep any today),
+// so hopefully we keep up the streak
+func importPathToFullPackage(importPath string) string {
```
lavalamp: sorry to ask for a change but can we call this "importPathToCanonical" or "canonicalizeImportPath" or something?

parser/parse.go
```diff
@@ -749,3 +753,14 @@ func (b *Builder) addVariable(u types.Universe, useName *types.Name, in *tc.Var)
 	out.Underlying = b.walkType(u, nil, in.Type())
 	return out
 }
+
+// importPathToFullPackage takes an import path and returns the actual package
+// it doesn't work for any folder actually called `vendor` (we don't dep any today),
```
lavalamp: add: but that's OK because this is not a legal go package name.

deads2k:
Updated name and added test

deads2k:
Looks like this may fix https://github.com/kubernetes/gengo/issues/30

thockin:
The reason I liked the "normalize on emit" model is because it guarantees that even if we end up with the same pkg vendored at 2 different levels, it will behave correctly.  Admittedly that is sort of a wacky thing to do but it is semantically legitimate and well-defined.  We're already playing a little loose wrt vendoring, but I'd rather not make that worse, if we can avoid it.

The specific example you pointed out with a const `conversionPackagePath` could be worked to call the normalize function no?  I guess that's going to be something of whack-a-mole, but we could doc our way out of it?

I'll review this again and see if I still hate it

deads2k:
> The specific example you pointed out with a const conversionPackagePath could be worked to call the normalize function no? I guess that's going to be something of whack-a-mole, but we could doc our way out of it?

I think that ends up too whack-a-mole across everyone using it.  It defies normal expectations and we probably have double digits of uses in our generators today.  the problem gets worse as external projects type godep this and use it with all the base APIs from kube as vendored.  See the servicecatalog problem and openshift would follow after the next rebase.

parser/parse.go
```diff
@@ -261,8 +262,9 @@ func (b *Builder) addDir(dir string, userRequested bool) error {
 	if err != nil {
 		return err
 	}
-	pkgPath := importPathString(buildPkg.ImportPath)
-	if dir != buildPkg.ImportPath {
+	canonicalPackage := canonicalizeImportPath(buildPkg.ImportPath)
+	pkgPath := importPathString(canonicalPackage)
```
thockin: What about making `canonicalizeImportPath` return type `importPathString`, and explicitly cast to string when needed?

parser/parse.go
```diff
@@ -749,3 +753,14 @@ func (b *Builder) addVariable(u types.Universe, useName *types.Name, in *tc.Var)
 	out.Underlying = b.walkType(u, nil, in.Type())
 	return out
 }
+
+// canonicalizeImportPath takes an import path and returns the actual package
+// it doesn't work for any folder actually called `vendor` (we don't dep any today),
```
thockin: I would change this comment that it doesn';t support nested vendoring, but since we don't do that and we don't know anyone who does that it's a reasonable asumption.

thockin:
On Tue, Jan 10, 2017 at 1:16 PM, David Eads <notifications@github.com> wrote:
>
> The specific example you pointed out with a const conversionPackagePath could be worked to call the normalize function no? I guess that's going to be something of whack-a-mole, but we could doc our way out of it?
>
> I think that ends up too whack-a-mole across everyone using it. It defies normal expectations and we probably have double digits of uses in our generators today. the problem gets worse as external projects type godep this and use it with all the base APIs from kube as vendored. See the servicecatalog problem and openshift would follow after the next rebase.

It's whack-a-mole but it is correct.  Package paths can't REALLY
canonicalize without context.  BUt we're already broken in the face of
that, so meh.

parser/parse.go
deads2k: > What about making canonicalizeImportPath return type importPathString, and explicitly cast to string when needed?

Removed four casts, added five.  I don't really care though.

deads2k:
Updated for @thockin's comments

deads2k:
To whoever feels merge-y first, can you get https://github.com/kubernetes/gengo/pull/29 as well.  It was noted "lgtm" here: https://github.com/kubernetes/gengo/pull/29#issuecomment-270761821

thockin:
LGTM

PR #32, opened 2017-01-11
deepcopy-gen: make deepcopy statically typed and without reflection

sttts:
Counterpart of https://github.com/kubernetes/kubernetes/pull/36412

TODO:
- [x] update documentation
- [x] add tests


thockin:
Are we going to bring this back for consideration in 1.7?  I've been letting it simmer and I am OK with generating methods, but I'd like to better understand the tag semantics here.  I've been tooling around with a better JSON generator, also generating methods.  I'd like to raise the bar wrt "generically useful outside of k8s" for gengo...

sttts:
@thockin I have nearly finished rebasing onto current master, want to pick that up and finish next week. Luckily it's not a rebase hell as expected.

I guess you refer to the new interface tag: `+k8s:deepcopy-gen:interfaces=k8s.io/kubernetes/runtime.Object`.

If you put this in front of a struct `Foo` the following rules apply:

1. if `Foo` has deepcopy generation enabled (via type opt-in or package wide opt-in), it will get an additional `DeepCopyObject() runtime.Object` method with the obvious implementation calling the normal deepcopy method.
2. if `Foo` opts out of deepcopy generation, it won't get the `DeepCopyObject()` either.
3. if another struct `Bar` embeds `Foo`, it inherits the `+k8s:deepcopy-gen:interfaces` tag (and transitively for further embeddings).

The whole mechanism is used to be able to deepcopy `runtime.Object`s and other interface types, without having to know the actual types of the objects behind the interface. In other words, this is a type-safe, still dynamic deepcopy dispatch, making us of Go's type system..

The 3rd, not so obvious rule has 2 important purposes:

- it allows us to avoid repeated tagging for all `TypeMeta` embedding types. By far most of our `runtime.Object`s have `TypeMeta` embedded. Hence, it is enough to have the following **once** in apimachinery:
```
// +k8s:deepcopy-gen=false
// +k8s:deepcopy-gen:interfaces=k8s.io/kubernetes/pkg/runtime.Object
type TypeMeta struct {
```
**Note** that `TypeMeta` itself has opted out of deepcopy. But our generators inline the deepcopy code anyway, so everything is fine.

- the **second purpose of the 3rd rule** is to avoid a potentially dangerous mistake: assume a type `Foo` that is tagged with `gen:interfaces=k8s.io/kubernetes/runtime.Object`. Another struct `Bar` embeds `Foo`. By Go's inheritence semantics, all funcs of `Foo` are also callable for `Bar`. Hence, one could call `Bar.DeepCopyObject()` which actually gives you only a copy of the embedded `Foo`, not of the whole of `Bar`. And worse: this is type-safe and the compiler won't tell you. With the third rule we enforce that `Bar` gets its own `DeepCopyObject()`.

sttts:
@thockin does this answer your question? If not, which aspect is open or unclear? We can also jump into a hangout if necessary.

sttts:
And in respect to being useful outside of k8s: only the tag names are kube specific, plus some special types like `Time`.

thockin:
On Fri, Mar 31, 2017 at 6:32 AM, Dr. Stefan Schimanski
<notifications@github.com> wrote:
>
> @thockin I have nearly finished rebasing onto current master, want to pick that up and finish next week. Luckily it's not a rebase hell as expected.
>
> I guess you refer to the new interface tag: +k8s:deepcopy-gen:interfaces=k8s.io/kubernetes/runtime.Object.
>
> If you put this in front of a struct Foo the following rules apply:
>
> if Foo has deepcopy generation enabled (via type opt-in or package wide opt-in), it will get an additional DeepCopyObject() runtime.Object method with the obvious implementation calling the normal deepcopy method.

Is this so it can satisfy some interface?  I don't understand the use
case for it...

> if Foo opts out of deepcopy generation, it won't get the DeepCopyObject() either.
> if another struct Bar embeds Foo, it inherits the +k8s:deepcopy-gen:interfaces tag (and transitively for further embeddings).

Not sure I get this one either.  It already inherits a method if it
embeds a type.

> The whole mechanism is used to be able to deepcopy runtime.Objects and other interface types, without having to know the actual types of the objects behind the interface. In other words, this is a type-safe, still dynamic deepcopy dispatch, making us of Go's type system..

I think I get it.  You want to add DeepCopyObject() to runtime.Object
interface, but it has to return Object, not RealType.  I get it, but
it's ugly.

WRT PR, maybe we can get a PR that has DeepCopy as a method without
the extra interfaces stuff, and then add that one on top?

Or maybe that one makes sense as different kube-specific generator?

> The 3rd, not so obvious rule has 2 important purposes:
>
> it allows us to avoid repeated tagging for all TypeMeta embedding types. By far most of our runtime.Objects have TypeMeta embedded. Hence, it is enough to have the following once in apimachinery:
>
> // +k8s:deepcopy-gen=false
> // +k8s:deepcopy-gen:interfaces=k8s.io/kubernetes/pkg/runtime.Object
> type TypeMeta struct {
>
> Note that TypeMeta itself has opted out of deepcopy. But our generators inline the deepcopy code anyway, so everything is fine.

That's a weird and hard to comprehend semantic.  At first glance, I'd
rather just .. not.. I don't have a better answer to be more
automatic, except that I hope we see a higher-level IDL which simply
says "this is an API type" and everything else is automatic.

> the second purpose of the 3rd rule is to avoid a potentially dangerous mistake: assume a type Foo that is tagged with gen:interfaces=k8s.io/kubernetes/runtime.Object. Another struct Bar embeds Foo. By Go's inheritence semantics, all funcs of Foo are also callable for Bar. Hence, one could call Bar.DeepCopyObject() which actually gives you only a copy of the embedded Foo, not of the whole of Bar. And worse: this is type-safe and the compiler won't tell you. With the third rule we enforce that Bar gets its own DeepCopyObject().

yeah, that's a funky and non-obvious case...

thockin:
I want to raise the bar.  It's mostly generic, but the semantics of the
tags and of gengo in general are still very kube-centric.  E.g. your
inheritance is very tailored to kubernetes.

sttts:
> if Foo has deepcopy generation enabled (via type opt-in or package wide opt-in), it will get an additional DeepCopyObject() runtime.Object method with the obvious implementation calling the normal deepcopy method.
> Is this so it can satisfy some interface?  I don't understand the use
> case for it...

The use-case is that you have interface references in objects and you want to deep-copy them. We have that at a number of places, e.g.:
- field and label selectors
- rbac rules
- embedded `runtime.Object`s through `runtime.RawExtension`

Interfaces hide the actual types of referenced objects such that you cannot statically call the `DeepCopy()` method. Moreover, you cannot add the `DeepCopy()` method to the interface because its type is not uniform (it's `DeepCopy() *Pod`, but you want `DeepCopy() runtime.Object`). Hence, – to avoid name conflicts – we add a mechanism to create `DeepCopyObject() runtime.Object`.

**Note** that this mechanism is not bound to any Kubernetes types, it completely generic.

> Not sure I get this one either.  It already inherits a method if it
> embeds a type.

Exactly. This is a problem if you do not override it (automatically, hence the tag rule #3): a call only deepcopies the embedded, not the whole object. Everything is type-correct, but still wrong. To avoid these kind of mistake, we inherit the tag (rule #3).

> I think I get it.  You want to add DeepCopyObject() to runtime.Object
> interface, but it has to return Object, not RealType.  I get it, but
> it's ugly.
>
> WRT PR, maybe we can get a PR that has DeepCopy as a method without
> the extra interfaces stuff, and then add that one on top?
> 
> Or maybe that one makes sense as different kube-specific generator?

I don't think we can and should split this. We have no way to copy interface types without using reflection again.

I might be biased, but I do not consider this mechanism ugly at all, but maybe the most powerful change in the whole PR: I discovered tons of cases where we do not provide deepcopy methods, but  fall back to the reflection based, super slow code path. In other word, this mechanism guarantees that a `runtime.Object` can be deepcopied, perfectly statically typed through Go interface semantics.

> That's a weird and hard to comprehend semantic.  At first glance, I'd
> rather just .. not.. I don't have a better answer to be more
> automatic, except that I hope we see a higher-level IDL which simply
> says "this is an API type" and everything else is automatic.

This is a bit tricky, I agree. And it is pure laziness. We can easily replace it with explicit tags, just by inserting them where we embed `TypeMeta`. Your "this is an API type" maps to `runtime.Object` in our type universe.

About being generic: other than the tag names: what exactly do you see as being Kubernetes specific right now? I can perfectly imagine to use this mechanism in any other type universe, it's not bound to the apimachinery one.

sttts:
> I think I get it. You want to add DeepCopyObject() to runtime.Object
> interface, but it has to return Object, not RealType. I get it, but
> it's ugly.
> 
> WRT PR, maybe we can get a PR that has DeepCopy as a method without
> the extra interfaces stuff, and then add that one on top?
> 
> Or maybe that one makes sense as different kube-specific generator?
> I don't think we can and should split this. We have no way to copy interface types without using reflection again.

@thockin thought about this again: we cannot get rid of the `DeepCopyObject() runtime.Object` easily, but we can drop the complicated rule #3 from this PR. I.e. we add expicit `+k8s:deepcopy-gen:interfaces` to all types with `TypeMeta` (easy to find with a grep). Then all the magic is gone and the semantics are easy.

sttts:
@thockin I have separated the interface promotion for embedded types from this PR into https://github.com/kubernetes/gengo/pull/44. The interface tag is now straight forward: it directly leads to generation of the `DeepCopyObject() runtime.Object` funcs without any semantical magic. I will update the kube counterpart PR.

smarterclayton:
Now that this is opt-in this is LGTM.  Can you rebase?

sttts:
@smarterclayton thanks. Will do.

sttts:
Targetting to get this ready as soon as possible after https://github.com/kubernetes/kubernetes/pull/44784 goes in.

sttts:
@smarterclayton kube PR is ready: #48544

As soon as this PR here is merged, I will update the vendoring commit in #48544

liggitt:
consuming PR has LGTM and tests are green except for godep check on this commit

smarterclayton:
LGTM.

PR #33, opened 2017-01-12
Preserve the filesystem location of a type

smarterclayton:
When dealing with vendored directories, we occasionally need to look at
the file path location. Copy the absPath for the package into the
package info so we have it available later in the gengo pipeline.

@deads2k @sttts @ncdc needed for kubernetes/kubernetes#39764


ncdc:
lgtm

PR #34, opened 2017-01-15
Tolerate vendored directories being scanned

smarterclayton:
Some packages may appear multiple times while traversing a tree (a
package vendored by two separate sub vendors). Preserve the first to
prevent types from being redefined.

In defaulter-gen, handle paths that contain /vendor/ by treating them as
top level packages.

Extracted from kubernetes/kubernetes#39916

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -226,8 +226,11 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 
 		var peerPkgs []string
 		if customArgs, ok := arguments.CustomArgs.(*CustomArgs); ok {
-			if len(customArgs.ExtraPeerDirs) > 0 {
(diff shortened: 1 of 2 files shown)


smarterclayton:
@sttts or @ncdc or @deads2k please review, this is the prerequisite for the DO NOT MERGE in the other repo

ncdc:
Would you ever expect an issue where the first package parsed is not the "right" one?

deads2k:
lgtm

deads2k:
This holds up https://github.com/kubernetes/kubernetes/pull/39916 which holds up https://github.com/kubernetes/kubernetes/pull/39894 which is needed to avoid confusion during re-vendoring.

smarterclayton:
@ncdc it could happen - at some point we'd probably need to add hooks to resolve package conflicts, or to error out if they are different.  Warning for now.

thockin:
How did this come up?  we're playing fast and loose with vendoring in a few places now, and I suspect it will come around to bite us eventually..

PR #37, opened 2017-01-23
Allow deepcopy to tolerate generating packages under /vendor/

smarterclayton:
Adjust the output path of the package to be within the vendor directory,
even if the source path is normalized to the short path (i.e.
k8s.io/apimachinery/pkg/apis/meta/v1 should be generated under
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/apis/meta/v1).

A future change to gengo should handle this for all generators rather
than requiring per generator changes.

Extracted from kubernetes/kubernetes#40178

@sttts @deads2k

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -184,10 +184,25 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 
 		if pkgNeedsGeneration {
 			glog.V(3).Infof("Package %q needs generation", i)
+			path := pkg.Path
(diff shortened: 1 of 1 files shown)


deads2k:
> A future change to gengo should handle this for all generators rather
> than requiring per generator changes.

Is that a thing you're looking at doing?

lgtm to fix the immediate problem.  This is being hit in master today: https://github.com/kubernetes/kubernetes/pull/36765

smarterclayton:
Maybe not today, but definitely I will steer someone at some point.

smarterclayton:
This change is consistent with our other short term "make gen in vendor possible" changes.  Merging on the green.

PR #38, opened 2017-01-29
WIP: Generated nested obj func.

k82cn:
part of https://github.com/kubernetes/kubernetes/pull/39713


k82cn:
/cc @smarterclayton , I'm updating it to support List and make it more  common.

It seems we can only use string compare for `InitContainer`, `Pod` and so on.

smarterclayton:
I don't know how useful this is to us over the long term.  Annotations as fields are temporary things that will be replaced in the future with generated machinery.  I don't really know that we want to maintain this if it's going to be used sparsely.

thockin:
I agree we probably don't need this.

PR #40, opened 2017-02-21
Fix lint failures in the deepcopy-gen

spzala:
deepcopy generated files. The total numbers of such lint failures are
around fifty errors. There are main three types of lint errors,
1. Generated functions name have underscore in them.
2. Generated functions are missing comment.
3. Some of the generated functions have unnecessary 'else' block for a given
   'if' loop.
For example, in the generated code, few of fifity some total failures are:
/zz_generated.deepcopy.go:60:6: don't use underscores in Go names;
/zz_generated.deepcopy.go:60:1: exported function should have comment or
be unexported

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -382,7 +382,7 @@ func (n *dcFnNamer) Name(t *types.Type) string {
 	if t.Name.Package == n.myPackage {
 		return "DeepCopy" + pubName
 	}
-	return fmt.Sprintf("%s.DeepCopy%s", n.tracker.LocalNameOf(t.Name.Package), pubName)
+	return fmt.Sprintf("%s.DeepCopy_%s", n.tracker.LocalNameOf(t.Name.Package), pubName)
 }
 
 func (g *genDeepCopy) Init(c *generator.Context, w io.Writer) error {
@@ -644,10 +644,10 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 			} else {
 				// TODO: don't depend on kubernetes code for this
 				sw.Do("if in.$.name$ != nil {\n", args)
-				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err == nil {\n", args)
-				sw.Do("out.$.name$ = *newVal.(*$.type|raw$)\n", args)
-				sw.Do("} else {\n", nil)
+				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err != nil {\n", args)
 				sw.Do("return err\n", nil)
+				sw.Do("} else {\n", nil)
+				sw.Do("out.$.name$ = *newVal.(*$.type|raw$)\n", args)
 				sw.Do("}\n", nil)
 				sw.Do("}\n", nil)
 			}
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -95,7 +95,7 @@ func extractTag(comments []string) *tagValue {
 func deepCopyNamer() *namer.NameStrategy {
 	return &namer.NameStrategy{
 		Join: func(pre string, in []string, post string) string {
-			return strings.Join(in, "_")
+			return strings.Join(in, "")
 		},
 		PrependPackageNames: 1,
 	}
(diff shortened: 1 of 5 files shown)


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -380,9 +380,9 @@ func (n *dcFnNamer) Name(t *types.Type) string {
 	pubName := n.public.Name(t)
 	n.tracker.AddType(t)
 	if t.Name.Package == n.myPackage {
-		return "DeepCopy_" + pubName
+		return "DeepCopy" + pubName
 	}
-	return fmt.Sprintf("%s.DeepCopy_%s", n.tracker.LocalNameOf(t.Name.Package), pubName)
+	return fmt.Sprintf("%s.DeepCopy%s", n.tracker.LocalNameOf(t.Name.Package), pubName)
```
smarterclayton: This is intentional.  I don't think we care what Lint thinks here.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -627,10 +628,10 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 			} else {
 				// Fall back on the slow-path and hope it works.
 				// TODO: don't depend on kubernetes code for this
-				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err != nil {\n", args)
-				sw.Do("return err\n", nil)
-				sw.Do("} else {\n", nil)
+				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err == nil {\n", args)
```
smarterclayton: If you're in here, invert the condition so that the return is in the if block.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -643,10 +644,10 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 			} else {
 				// TODO: don't depend on kubernetes code for this
 				sw.Do("if in.$.name$ != nil {\n", args)
-				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err != nil {\n", args)
-				sw.Do("return err\n", nil)
-				sw.Do("} else {\n", nil)
+				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err == nil {\n", args)
```
smarterclayton: Invert the condition.

examples/deepcopy-gen/generators/deepcopy.go
smarterclayton: Ah, I see that this is creating locally scoped errors.  This is still ugly, but I guess I can live with it.

spzala: 👍 :) yes, and thank you so much for noticing it.

examples/deepcopy-gen/generators/deepcopy.go
spzala: 👍 sure, will do.

examples/deepcopy-gen/generators/deepcopy.go
spzala: @smarterclayton sure, sounds good. Thanks!!

duglin: If we don't fix all lint issues then people who generate code need to go out of their way to tell lint to skip these files - which is really annoying. Any reason why we don't want to follow lint's guidance?

duglin: and yes lint is just suggestions but people run it, including kube, so we might as well follow thru on that thought.

spzala: @duglin yes that's a valid question to me and in fact as a newbie to kube I thought of the same question to ask to @smarterclayton for curiosity, for future reference and to share with stack holders of the original requirement. But then after undoing this particular change and testing it with lint, I noticed that lint didn't complain. The lint violation is with the actual name of function (LOC 383). Thanks!

spzala: @duglin @smarterclayton as you might already noticed, the lint errors clutters the test output and as @duglin said unless we skip it, the cluttered test output can be a painful reading.

spzala:
@smarterclayton just a gentle reminder on review :-) Thanks!

spzala:
@smarterclayton Hi, since this patch is around for some long time, a gentle review reminder :-) Please let me know if you have any question for me. Thanks!!

sttts:
@spzala can you please rebase and I will take a look.

spzala:
@sttts thank you so much but I am closing this PR per earlier discussion with broader audience that we want to keep underscores in the func name. Few of other lint issues are fixed under diff PRs as mentioned here https://github.com/kubernetes/kubernetes/pull/41764

PR #41, opened 2017-03-02
Fix lint failures in the defaulter generated files

spzala:
defaulter generated files. There are main two types of lint errors,
1. Generated functions name have underscore in them.
2. Generated functions are missing comment.

For example, in the generated code, some of the failures are:
zz_generated.defaults.go:53:6: don't use underscores in Go names…
zz_generated.defaults.go:38:1: exported function SetObjectDefaults_Binding
should have comment or be unexported..

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -55,7 +55,7 @@ func checkTag(comments []string, require ...string) bool {
 
 func defaultFnNamer() *namer.NameStrategy {
 	return &namer.NameStrategy{
-		Prefix: "SetDefaults_",
+		Prefix: "SetDefaults",
 		Join: func(pre string, in []string, post string) string {
 			return pre + strings.Join(in, "_") + post
 		},
@@ -64,7 +64,7 @@ func defaultFnNamer() *namer.NameStrategy {
 
 func objectDefaultFnNamer() *namer.NameStrategy {
 	return &namer.NameStrategy{
-		Prefix: "SetObjectDefaults_",
+		Prefix: "SetObjectDefaults",
 		Join: func(pre string, in []string, post string) string {
 			return pre + strings.Join(in, "_") + post
 		},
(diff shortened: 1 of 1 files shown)


spzala:
@smarterclayton Hi Clayton, found some more lint issues and I think this seems pretty much it in gengo library lint related changes needed per our testing. In order to keep it clean for review and future reference, I have opened this as a separate PR then other PR which is for deepcopy generated with many changes. Hope makes sense. Thanks!!

spzala:
@smarterclayton a gentle reminder for review :-) Thanks!!  /cc @duglin

thockin:
This is generated code - the stylistic naming is on purpose.

spzala:
@thockin Hello Tim, thanks for the quick reply. So I agree but for projects (e.g. service-catalog) trying to be fully compliant with lint, requires to  skip generated code by modifying Make/scripts. We had a good discussion on a similar topic under this PR https://github.com/kubernetes/kubernetes/pull/42781 which is now merged, if you can take a quick look for more detail specially the thread of Mar 9. Unless and until all generated code are compliant with lint, we need to keep handling it or we can make such changes as in this PR and get rid of special checks for generated code. Thanks!! /cc @duglin

smarterclayton:
Please don't remove the underscore, but it's ok to add the godoc.

spzala:
@smarterclayton Hi Clayton, I know but without underscore we will face the same issue of handing lint by mean of skipping it. But for now, I have created https://github.com/kubernetes/gengo/pull/45 which doesn't remove underscore and takes care of two other failures which we discussed earlier. I believe it's good idea to have them in and we can have some more discussion on underscore related changes. Hope makes sense. Thanks!!

spzala:
I am closing this PR per earlier discussion with broader audience that we want to keep underscore in the func names. Few of other lint issues are fixed under diff PRs as mentioned here https://github.com/kubernetes/kubernetes/pull/41764

PR #42, opened 2017-03-10
Fix bug in generating vars / consts

pwittrock:


lavalamp:
LGTM - failure is at tip, looks like go parser line number identification must have changed.

PR #44, opened 2017-04-10
deepcopy: add interface promotion for embedded structs

sttts:
Based on https://github.com/kubernetes/gengo/pull/32

```golang
+// In addition, if a type T embeds another type T' which implements deepcopy for
 +// an interface, this property if promoted to T and the respecitve DeepCopyInterfaceName
 +// is also generated, i.e. those +k8s:deepcopy-gen:interfaces tags are applied
 +// transitively if types are anonymously embedded.
 +//
 +// Note: types can be tagged with +k8s:deepcopy-gen:interfaces and +k8s:deepcopy-gen=false
 +// at the same time. Then no methods are generated, but this property can be promoted
 +// to other types via embeddings (and those will get the auto-generated DeepCopyInterfaceName
 +// methods). E.g. TypeMeta is such a type in Kubernetes. This makes it easy to apply the
 +// tag one struct only, and all types embedding that struct will implement the given
 +// interfaces. No manual tagging is neccessary then.
 +//
```


thockin:
dredging this up - we added some much simpler output_tests that should highlight the changes you're making in each commit, once you rebase and add a testcase for interfaces..

sttts:
@thockin I have rebased this one and https://github.com/kubernetes/gengo/pull/32. What do I have to do to push this forward? From my side, I need a lgtm here or on #32 first for the concept and the generator and then will rebase https://github.com/kubernetes/kubernetes/pull/36412 accordingly again.

sttts:
#32 has merged. Closing this one.

PR #45, opened 2017-04-19
Fix some of the straightforward lint failures in the deepcopy-gen

spzala:
deepcopy generated files. The total numbers of such lint failures are
around fifty errors. There are main three types of lint errors,
1. Generated public functions are missing comment.
2. Some of the generated functions have unnecessary 'else' block for a given
   'if' loop.
The third type of lint failure requires removing underscore from function
names, which is not modified in this PR as it requires further discussion
and consensus.
For example, in the generated code, few of fifity some total failures are:
/zz_generated.deepcopy.go:60:1: exported function should have comment or
be unexported


spzala:
@smarterclayton Hi, I am working on this PR (no underscore related changes for link in this one) but even though with just a one liner change, Travis is failing :( .. I could run test.sh locally fine. Can you please take a look? Thanks!

spzala:
@thockin  Thank you much Tim!! (FYI, I initially added one more lint related change under this PR but reverted because Travis was failing so wanted to try minimal changes to debug the issue. I will add those changes in another PR.)

PR #46, opened 2017-04-20
Fix if loop related lint failures in the deepcopy-gen

spzala:
The changes in this patch takes care of lint failures due to unreachable
else statements. For example one of the failures is,

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -523,11 +523,12 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 				sw.Do("}\n", nil)
 				sw.Do("(*out)[key] = *newVal\n", nil)
 			} else {
-				sw.Do("if newVal, err := c.DeepCopy(&val); err != nil {\n", nil)
+				sw.Do("newVal, err := c.DeepCopy(&val)\n", nil)
+				sw.Do("if err != nil {\n", nil)
 				sw.Do("return err\n", nil)
-				sw.Do("} else {\n", nil)
-				sw.Do("(*out)[key] = *newVal.(*$.|raw$)\n", t.Elem)
 				sw.Do("}\n", nil)
+				sw.Do("(*out)[key] = *newVal.(*$.|raw$)\n", t.Elem)
+				sw.Do("\n", nil)
 			}
 			sw.Do("}\n", nil)
 		}
@@ -564,11 +565,12 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
 			sw.Do("return err\n", nil)
 			sw.Do("}\n", nil)
 		} else {
-			sw.Do("if newVal, err := c.DeepCopy(&(*in)[i]); err != nil {\n", nil)
+			sw.Do("newVal, err := c.DeepCopy(&(*in)[i])\n", nil)
+			sw.Do("if err != nil {\n", nil)
 			sw.Do("return err\n", nil)
-			sw.Do("} else {\n", nil)
-			sw.Do("(*out)[i] = *newVal.(*$.|raw$)\n", t.Elem)
 			sw.Do("}\n", nil)
+			sw.Do("(*out)[i] = *newVal.(*$.|raw$)\n", t.Elem)
+			sw.Do("\n", nil)
 		}
 		sw.Do("}\n", nil)
 	}
@@ -628,11 +630,12 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 			} else {
 				// Fall back on the slow-path and hope it works.
 				// TODO: don't depend on kubernetes code for this
-				sw.Do("if newVal, err := c.DeepCopy(&in.$.name$); err != nil {\n", args)
+				sw.Do("newVal, err := c.DeepCopy(&in.$.name$)\n", args)
+				sw.Do("if err != nil {\n", nil)
 				sw.Do("return err\n", nil)
-				sw.Do("} else {\n", nil)
-				sw.Do("out.$.name$ = *newVal.(*$.type|raw$)\n", args)
 				sw.Do("}\n", nil)
+				sw.Do("out.$.name$ = *newVal.(*$.type|raw$)\n", args)
+				sw.Do("\n", nil)
 			}
 		default:
 			// Interfaces, Arrays, and other Kinds we don't understand.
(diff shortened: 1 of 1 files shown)


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -523,10 +523,10 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 				sw.Do("}\n", nil)
 				sw.Do("(*out)[key] = *newVal\n", nil)
 			} else {
-				sw.Do("if newVal, err := c.DeepCopy(&val); err != nil {\n", nil)
-				sw.Do("return err\n", nil)
-				sw.Do("} else {\n", nil)
+				sw.Do("if newVal, err := c.DeepCopy(&val); err == nil {\n", nil)
```
thockin: It's much more idiomatic to test err != nil - why change that?  I had a hard time realizing what you changed because my eyes just skip the pattern.  I'd rather see a test for != nil and a `return nil` in the success path

spzala: @thockin Hi Tim, lol, sorry and I agree. Honestly I didn't like it myself too but I notice use of == nil in few other places and I was getting locally scoped error with a different solution I tried few weeks back (part of an old PR which didn't go further because of underscore related changes). But seems like a possible better solution is working now, I will update the changes. Thanks!

thockin:
This is an annoying and stupid lint error.  The `if x := func(); x { ... } else { ... }` is idiomatic and should be encouranged when x is not used outside the condition blocks.  

sigh

thockin:
change LGTM but fails travis

thockin:
restarted travis - 1/3 failed

spzala:
@thockin I know :( and thanks much!! It seems like no straightforward error with Travis? So I guess may be one more restart on a weekday?

thockin:
It is failing against Go tip, consistently.  I am on mobile, so can't investigate.

```
--- FAIL: TestSnippetWriter (0.00s)
	snippet_writer_test.go:85: Expected "snippet_writer_test.go:78" but didn't find it in "template: /home/travis/gopath/src/k8s.io/gengo/generator/snippet_writer_test.go:77:1: unclosed action"
FAIL
FAIL	k8s.io/gengo/generator	0.006s
```

spzala:
@thockin Thanks Tim and no rush!! I noticed that it was the same failure even in previous patch with no code change (i.e. only  a comment was added https://github.com/kubernetes/gengo/pull/45 ). Seems strange :(

thockin:
It looks like something change in upstream Go wrt line numbers.

Can you fold in a change (separate commit) to `generator/snippet_writer_test.go` to remove the literal line number? and see if that passes?

spzala:
@thockin Hi Tim, very nice, that seems did the trick .. I created this one https://github.com/kubernetes/gengo/pull/48 and Travis's happy. Thanks!!

thockin:
rebase and ping me

spzala:
@thockin Hi Tim, I tried pinging you on slack kube-dev couple times but without much luck of talking to you :-) Just an update that this PR is passing Travis. Thanks!!

thockin:
I just restarted travis - I added tests that this will have to update.
Resync to master and look at examples/deepcopy/output_tests

thockin:
Specifically, run `make test` in that dir.

spzala:
Thanks Tim @thockin

spzala:
@thockin is there a way I can restart Travis job by adding a comment or something like that? (I did rebase and updated PR but seems like latest Travis failure is due to incomplete build job and probably need a restart) Thanks!

thockin:
like that

spzala:
Ahhh.. interesting.. Close the PR and ReOpen. Different but that's cool :-) Thanks @thockin !!

spzala:
@thockin Yay :-) it passed now.

thockin:
Curiously, no test cases exercize this.  I'm going to merge it because that's not your fault, but if you want to make me extra happy, maybe you could throw together super-minimal output_test cases that expose this change?

spzala:
@thockin :-) Thanks Tim! And yes sure, I would work on it. I am traveling on business for a week and preparing for my speaking session etc but I believe there is no super rush with it right? I will be working on soon I can.  Thanks much!!

thockin:
Also it should be SUPER EASY.  Copy on of the directories, edit doc.g to
have a type that highlights this case, run `make test`

spzala:
Awesome, thanks @thockin !!

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -523,11 +523,12 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 				sw.Do("}\n", nil)
 				sw.Do("(*out)[key] = *newVal\n", nil)
 			} else {
-				sw.Do("if newVal, err := c.DeepCopy(&val); err != nil {\n", nil)
+				sw.Do("newVal, err := c.DeepCopy(&val)\n", nil)
```
liggitt: this produces code that will not compile

see https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/45797/pull-kubernetes-bazel/20933/

```
func DeepCopy_apiextensions_CustomResource(in interface{}, out interface{}, c *conversion.Cloner) error {
	{
		in := in.(*CustomResource)
		out := out.(*CustomResource)
		*out = *in
		newVal, err := c.DeepCopy(&in.ObjectMeta)
		if err != nil {
			return err
		}
		out.ObjectMeta = *newVal.(*v1.ObjectMeta)

		newVal, err := c.DeepCopy(&in.Spec)
		if err != nil {
			return err
		}
		out.Spec = *newVal.(*CustomResourceSpec)

		newVal, err := c.DeepCopy(&in.Status)
		if err != nil {
			return err
		}
		out.Status = *newVal.(*CustomResourceStatus)

		return nil
	}
}
```

liggitt: opened https://github.com/kubernetes/gengo/pull/56 to revert so we can pull master (we need to bump as a prereq of https://github.com/kubernetes/kubernetes/pull/45294)... can revisit this if we decide we want to jump through scoping or variable name hoops to avoid this lint error (I personally don't think it's worth it)

spzala: @liggitt thanks for fixing it Jordan. My bad, in my manual verification of generated code and build I apparently didn't run into it. I guess Gengo Travis passed because the method with nil return is not in Gengo but K8s. @thockin was quick in seeing a need for tests when lint related these changes were merged and kindly created new tests but those tests still missed few scenarios. I have also create a follow up PR (https://github.com/kubernetes/gengo/pull/51) to add new tests.  I value your opinion on spending time for lint may not worth :-) but if we can fix existing code on lint and going forward have lint complied code then that will reduce skipping lint check and produce more clean code. Thanks!!

liggitt: lint is a moving target, and (as in this case) sometimes encourages non-idiomatic code. I don't think jumping through hoops to silence it is a good i dea.

liggitt: > I have also create a follow up PR (#51) to add new tests.

I don't think that test exercises this case, either... this error was caused by multiple fields in the deep-copied type resulting in newVal being redeclared multiple times in the same scope.

spzala: @liggitt about lint agree as long as we try to catch it manually but at the same time I guess what can be worth is to automate test process against lint (latest) as part of the build (something similar to what's done with PEP 8 in projects like OpenStack), that should be a real solution.

spzala: @liggitt  for #51 do you think whatever is added in the PR are valuable tests for those specific scenarios and good be have? Thanks!

PR #47, opened 2017-04-24
Udpates deepcopy-gen to allow input types being defined in a different pacakge

caesarxuchao:
api/v1/types.go is moved to staging in https://github.com/kubernetes/kubernetes/pull/44784, and will be moved to its own repo in the future. This breaks deepcopy-gen's assumption that types are defined at the same location as the output deepcopy functions. This PR adds required flexibility to deepcopy-gen.

One can specify the location input types with `types` key, the value will be the relative dir to where the types are defined (see https://github.com/caesarxuchao/kubernetes/commit/02ea1c78c4cf36f430fed204d200f9389f80b764):
```golang
+k8s:deepcopy-gen=package,register,types=../../../vendor/k8s.io/api/core/v1
```

The example output is at https://github.com/caesarxuchao/kubernetes/commit/d4afd56fb996d4c0dc9fe019ee51cd1906b5e24e. It's effectively the same as the manual changes i made to deepcopy functions in https://github.com/kubernetes/kubernetes/pull/44784.


caesarxuchao:
@lavalamp can you review it? Thanks.

thockin:
Changes here are becoming impossible to review.  We need a set out output tests that clearly show the implications.  I REALLY REALLY want this to get simpler, not to add more tags.

thockin:
Specifically we have a `test` directory here which should have a collection of test cases, each of which is small enough to manually inspect the generated code.

caesarxuchao:
@thockin

I agree with adding a test in this repo. I'll send another PR for that and then rebase this one.

Regarding tags, do you have any suggestion? IMO we are adding one degree of freedom here so it's reasonable to add a tag.

thockin:
I don't understand what happens without the tag - do you have a commit that shows the problem the tag fixes?

caesarxuchao:
In #44784 i moved pkg/api/v1/types.go to k8s.io/api/core/v1/types.go. Then I **manually** fixed all the generated files. This PR updates the generator so that the generated code will be the same as my manual fix in #44784. 
I listed some diff between the current generated code and the expected generated code (i.e., with the extra `types=../../../vendor/k8s.io/api/core/v1` tag)
https://github.com/kubernetes/kubernetes/pull/44784/files#diff-eb39927a03508dfb6dff248de30d9385R26
https://github.com/kubernetes/kubernetes/pull/44784/files#diff-70814d2013379c298777f661a84a1725R41
https://github.com/kubernetes/kubernetes/pull/44784/files#diff-70814d2013379c298777f661a84a1725R222
https://github.com/kubernetes/kubernetes/pull/44784/files#diff-f81b6a0b83b2a3011788136a9e0722a3R560

lavalamp:
I am confused about why this needs changes in the deep copy generator (as opposed to just running the generator twice). I would have expected that only the conversion generator needs to know about both the old and new types at the same time?

caesarxuchao:
Given that https://github.com/kubernetes/kubernetes/pull/36412 is going to make deep copies a method of the types, i'll just move deep copy functions to the type repo.

thockin:
I see that this change is effectively a no-op, but I don't understand what
happens if you DON'T have this change...

PR #49, opened 2017-04-25
remove init func from deepcopy.go

caesarxuchao:
Removing these lines: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/zz_generated.deepcopy.go#L31-L33

They should be registered here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/register.go#L37

@lavalamp can you take a look? Thanks.


thockin:
Given the work to make these into methods - is this worth it?

caesarxuchao:
I don't think https://github.com/kubernetes/kubernetes/pull/36412 is getting merged any soon, so i want to get this one and https://github.com/kubernetes/kubernetes/pull/44888 in to unblock the move of api types.

thockin:
needs rebase and update tests

caesarxuchao:
Now #44888 doesn't need changes in zz_generated.Deepcopy.go. Closing.

PR #51, opened 2017-04-30
Add new test for deepcopy

spzala:
Add new test convering servicecatalog types related to changes made under
https://github.com/kubernetes/gengo/pull/46


thockin:
I appreciate the followup.  Can we narrow this down and rename it to reflect the type that is being demonstrated.  Specifically, a struct with a single member field that triggers the rendered code - that way each output test is trivial to look at and decide if changes are correct.  It might mean 3 or 4 individual new dirs..

spzala:
@thockin - yrw and sure, that makes sense. I will update the changes. Thanks!!

spzala:
@thockin - Hi Tim, as you could tell I am new to Go, but have tried to identify new types and added tests for them. Hope that makes sense. Thanks!!

examples/deepcopy-gen/output_tests/interfaces/doc.go
```diff
@@ -0,0 +1,28 @@
...
+
+// This is a test package.
+package interfaces
+
+type Inter interface {
+  function() float64
+}
+
+type Ttest struct {
+  i []Inter
```
thockin: I think this has to be named `I` - non-private name.

examples/deepcopy-gen/output_tests/interfaces/doc.go
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// +k8s:deepcopy-gen=package
+
+// This is a test package.
+package interfaces
+
+type Inter interface {
```
thockin: What does `Inter` mean?

examples/deepcopy-gen/output_tests/pointer/zz_generated.go
```diff
@@ -0,0 +1,55 @@
...
+func DeepCopy_pointer_Ttest(in interface{}, out interface{}, c *conversion.Cloner) error {
+	{
+		in := in.(*Ttest)
+		out := out.(*Ttest)
+		*out = *in
+		if in.Types != nil {
+			in, out := &in.Types, &out.Types
+			*out = make(map[string]*Ttest)
+			for key, val := range *in {
+				newVal, err := c.DeepCopy(&val)
```
thockin: Why do we expect this to hit the Cloner path?

thockin:
LGTM!

PR #52, opened 2017-05-11
Generate defaulters for aliased types correctly

liggitt:
Needed by https://github.com/kubernetes/kubernetes/pull/45294 (that PR already includes this change, so you can see it "in action"... it's required in order to keep generating the recursive defaulting functions through aliased slice types)

Otherwise, aliased slice types don't detect registered defaulters on the element types


liggitt:
@smarterclayton PTAL

PR #53, opened 2017-05-12
fix a bug in gengo that parses comments in doc.go multiple times

caesarxuchao:
findTypesIn might be called multiple times and thus the content of doc.go might be recorded in the universe multiple time.

@lavalamp @thockin

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -458,6 +458,9 @@ func (b *Builder) findTypesIn(pkgPath importPathString, u *types.Universe) error
 	for _, f := range b.parsed[pkgPath] {
 		if strings.HasSuffix(f.name, "/doc.go") {
 			tp := u.Package(string(pkgPath))
+			// findTypesIn might be called multiple times. Clean up tp.Comments
+			// to avoid repeatedly fill same comments to it.
+			tp.Comments = []string{}
 			for i := range f.file.Comments {
 				tp.Comments = append(tp.Comments, splitLines(f.file.Comments[i].Text())...)
 			}
--- a/types/types.go
+++ b/types/types.go
@@ -101,10 +101,11 @@ type Package struct {
 	// 'package x' line.
 	Name string
 
-	// DocComments from doc.go, if any.
+	// The comment right above the package declaration in doc.go, if any.
 	DocComments []string
 
-	// Comments from doc.go, if any.
+	// All comments from doc.go, if any.
+	// TODO: remove Comments and use DocComments everywhere.
 	Comments []string
 
 	// Types within this package, indexed by their name (*not* including


caesarxuchao:

mbohlool:
LGTM. I have no permission to merge or label or anything :)

lavalamp:
@thockin @wojtek-t blame tells me one of you ought to know what is going on here. Why is there a Comments AND a DocComments, both with the same comment, yet this code looks the way it does?

caesarxuchao:
> Comments AND a DocComments

@lavalamp i added a commit to clarify the difference.

parser/parse.go
```diff
@@ -458,6 +458,7 @@ func (b *Builder) findTypesIn(pkgPath importPathString, u *types.Universe) error
 	for _, f := range b.parsed[pkgPath] {
 		if strings.HasSuffix(f.name, "/doc.go") {
 			tp := u.Package(string(pkgPath))
+			tp.Comments = []string{}
```
wojtek-t: Why do we need this line? You can append to "nil" slice, so I don't fully understand this line.

caesarxuchao: The following three lines is a loop that append to tp.Comments. What do you mean by appending to "nil" slice?

types/types.go
```diff
@@ -101,10 +101,10 @@ type Package struct {
 	// 'package x' line.
 	Name string
 
-	// DocComments from doc.go, if any.
+	// The comment right above the package declaration in doc.go, if any.
 	DocComments []string
 
-	// Comments from doc.go, if any.
+	// All comments from doc.go, if any.
```
lavalamp: This doesn't make any sense :) doc.go shouldn't have any other comments if it is following go style.

caesarxuchao: Sure. I was trying to interpret what's the original intention. I can remove one of them.

caesarxuchao: Removing one of Comments/DocComment will be a backward incompatible change, do you want me to do that?

lavalamp:
That's why I'm trying to get to the bottom of why it is like this.

parser/parse.go
thockin: if tp.Comments == nil, it is still valid to append to it.  This initialization line is effectively a no-op, unless there were contents that you wanted to throw away?

lavalamp:
The non-obvious thing is that this code is called multiple times on the
same thing (which may itself be the bug, actually).

types/types.go
thockin: Our distinction of DocComments is a little weird, anyway.  doc.go is a convention, but not a requirement.

For example:  https://godoc.org/k8s.io/gengo/args

There's no doc.go there, and yet it extracts the package comments correctly.

caesarxuchao: > For example: https://godoc.org/k8s.io/gengo/args There's no doc.go there, and yet it extracts the package comments correctly.

That's not what DocComments is doing, according to my experiment, it just has "The comment right above the package declaration in doc.go, if any."

`Comments` contains "All comments from doc.go, if any.", and I agree with @lavalamp it doesn't make sense.

For backwards compatibility, i think the current PR is the best.

parser/parse.go
lavalamp: Please include a // TODO explaining this line since if neither thockin nor wojtek-t figured out what was here for it must be pretty mysterious.

Then I'll hold my nose and lgtm.

I feel like taking a week off just to fix all the things that have been done to this library :(

thockin:
@lavalamp it's yours to merge when you're happy.  Not sure why it parses multiple times, but in playing with this I have had many things I have wanted to fix - especially around comments, doc-comments, second-closest comments, etc.

types/types.go
```diff
@@ -101,10 +101,11 @@ type Package struct {
...
+	// TODO: remove Comments and use DocComments everywhere.
```
caesarxuchao: @lavalamp i added a TODO here. PTAL.

lavalamp:
LGTM

PR #54, opened 2017-05-12
Default-gen changes for moving API types to k8s.io/api

caesarxuchao:
Currently, the `k8s:defaulter-gen` tag is in the `k8s.io/kubernetes/pkg/apis/<group>/<version>/doc.go`, co-located with the input (API types) and the output (default func) of default-gen.

After https://github.com/kubernetes/kubernetes/issues/44065 is implemented, the input/output/tag are not going to be in the same package:
* The input, the API types will be moved to k8s.io/api;
* The output, the generated default funcs, will stay in k8s.io/kubernetes, because only apiserver should ever need them; the tag (i.e., doc.go) will also stay in k8s.io/kubernetes.

This PR adds the `k8s:defaulter-gen-input` tag, so that the API types can be supplied from a different package.

@lavalamp @thockin


caesarxuchao:
I'll add some tests.

caesarxuchao:

mbohlool:
LGTM, do you want to add some tests before merging this?

caesarxuchao:
Yes, after the other is merged.

caesarxuchao:
The tests PR is not going anywhere, can we get this merged before code freeze? It's a noop currently until i set up the codegen tags in https://github.com/kubernetes/kubernetes/pull/44784.

parser/parse.go
```diff
@@ -260,6 +260,18 @@ func (b *Builder) AddDirTo(dir string, u *types.Universe) error {
 	return b.findTypesIn(canonicalizeImportPath(b.buildPackages[dir].ImportPath), u)
 }
 
+// AddDirToAndReturnAddedPath is the same as AddDirTo except that it also
+// returns the path of the added package.
+func (b *Builder) AddDirToAndReturnAddedPath(dir string, u *types.Universe) (string, error) {
```
lavalamp: Instead of doing this, add a method that canonicalizes a path and returns a *Package object.

parser/parse.go
```diff
@@ -260,6 +260,26 @@ func (b *Builder) AddDirTo(dir string, u *types.Universe) error {
 	return b.findTypesIn(canonicalizeImportPath(b.buildPackages[dir].ImportPath), u)
 }
 
+func (b *Builder) CanonicalPath(dir string) (string, error) {
```
caesarxuchao: Do you mean adding a method like this?

lavalamp: how about `DirectoryToPackage(dir string, u *types.Universe) (*types.Package, error)`

parser/parse.go
```diff
@@ -260,6 +260,26 @@ func (b *Builder) AddDirTo(dir string, u *types.Universe) error {
...
+	p, ok := b.buildPackages[dir]
+	if !ok {
+		return "", fmt.Errorf("buildPackages doesn't have %s", dir)
+	}
+	return nil, canonicalizeImportPath(b.buildPackages[dir].ImportPath)
+}
+
+// AddDirToAndReturnAddedPath is the same as AddDirTo except that it also
+// returns the path of the added package.
+func (b *Builder) AddDirToAndReturnAddedPath(dir string, u *types.Universe) (string, error) {
```
lavalamp: and remove this function

parser/parse.go
caesarxuchao: Returning the `package` is not going to help me. I need the `canonicalizeImportPath(package.ImportPath)`.

parser/parse.go
```diff
@@ -260,6 +260,16 @@ func (b *Builder) AddDirTo(dir string, u *types.Universe) error {
 	return b.findTypesIn(canonicalizeImportPath(b.buildPackages[dir].ImportPath), u)
 }
 
+// DirectoryToPackage returns the package path given a dir. dir must be already
+// parsed.
+func (b *Builder) DirectoryToPackagePath(dir string) (string, error) {
```
caesarxuchao: How about this? 

Returning Package is not going to help me, because what I need is canonicalizeImportPath(package.ImportPath).

I actually feel my original code is better. This version makes the generator call context.AddDir and context.DirectoryToPackagePath, while my version only calls context.AddDirAndReturnAddedPkgPath.

caesarxuchao: I still think my first version is better. I can give the method a better name, maybe AddDirectory (just to be different from AddDir). WDYT?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -271,8 +281,33 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
...
+			err := context.AddDir(inputDir)
+			if err != nil {
+				glog.Fatalf("cannot import package %s", inputTags[0])
+			}
+			inputPkgPath, err := context.DirectoryToPackagePath(inputDir)
+			if err != nil {
+				glog.Fatalf("cannot translate dir %s to package path package %s", inputDir)
+			}
+			var ok bool
+			typesPkg, ok = context.Universe[inputPkgPath]
```
lavalamp: Isn't this just getting the package? Why do you need its path?

caesarxuchao: context.Universe[inputPkgPath] returns k8s.io/gengo/types#Package.
`context.AddDirTo` calls builder to parse the source code, which only knows go/build#Package and go/types#Package.

The translation from directory to canonical path is done in builder, i can add a method to it to export the functionality, and then export this method via context. But I think that is worse than just return the package path via `AddDirTo`.

caesarxuchao:
Updated. PTAL. Thanks.

parser/parse.go
```diff
@@ -246,11 +246,7 @@ func (b *Builder) AddDirRecursive(dir string) error {
 	return nil
 }
 
-// AddDirTo adds an entire directory to a given Universe. Unlike AddDir, this
-// processes the package immediately, which makes it safe to use from within a
-// generator (rather than just at init time. 'dir' must be a single go package.
-// GOPATH, GOROOT, and the location of your go binary (`which go`) will all be
-// searched if dir doesn't literally resolve.
+// Deprecated. Please use AddDirectoryTo.
```
lavalamp: Please leave the former godoc comment and just append this note :)

generator/generator.go
```diff
@@ -203,9 +203,14 @@ func NewContext(b *parser.Builder, nameSystems namer.NameSystems, canonicalOrder
 	return c, nil
 }
 
-// AddDir adds a Go package to the context. The specified path must be a single
-// go package import path.  GOPATH, GOROOT, and the location of your go binary
-// (`which go`) will all be searched, in the normal Go fashion.
+// Deprecated. Please use AddDirectory.
```
lavalamp: Same

lavalamp:
LGTM

PR #55, opened 2017-05-12
add a test for default-gen output

caesarxuchao:
cc @lavalamp @thockin


caesarxuchao:

examples/defaulter-gen/Makefile
```diff
@@ -0,0 +1,15 @@
+TOOL=defaulter-gen
+
+test:
+	@if ! git diff --quiet HEAD; then \
+	    echo "FAIL: git client is not clean"; \
+	    false; \
+	fi
+	@go build -o /tmp/$(TOOL)
+	@PKGS=$$(go list ./output_tests/...  | paste -sd' '); \
+	/tmp/$(TOOL) --logtostderr --v=6 -i $$(echo $$PKGS | sed 's/ /,/g') -O zz_generated
```
mbohlool: so the paste command above is adding space between package names and sed here convert those spaces to comma, right? why not just use `paste -sd','`?

examples/defaulter-gen/output_tests/wholepkg/vendor/k8s.io/apimachinery/pkg/runtime/mock.go
```diff
@@ -0,0 +1,22 @@
+/*
```
mbohlool: Why do you want the generated file to be buildable? is the test only making sure that the output is what we expecting?

caesarxuchao: The travis builds the entire repo.

examples/defaulter-gen/Makefile
caesarxuchao: I copy-pasted it from https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/authorization/v1/types.go#L27. I can update both if you think it's necessary.

mbohlool: The link is referring to the wrong file? Anyway I think you can just fix it here if you like. It is not necessary though.

examples/defaulter-gen/output_tests/wholepkg/vendor/k8s.io/apimachinery/pkg/runtime/mock.go
caesarxuchao: I can try to change .travis but that's more nasty IMHO :)

mbohlool: In that case, I suggest you rename the expected file zz_generated.go to zz_generated.expected, add the name zz_generated.go to git_ignore of that folder, and do a `diff zz_generated.expected zz_generated.go` instead of `git diff ...`.

examples/defaulter-gen/Makefile
caesarxuchao: Sorry, https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/Makefile#L9

caesarxuchao:
@mbohlool updated, ptal. Thanks!

mbohlool:
LGTM

mbohlool:
But it look like travis is still compiling it :( maybe remove the file after you checked the diff?

examples/defaulter-gen/Makefile
```diff
@@ -0,0 +1,15 @@
+TOOL=defaulter-gen
+
+test:
+	@if ! git diff --quiet HEAD; then \
+	    echo "FAIL: git client is not clean"; \
+	    false; \
+	fi
+	@go build -o /tmp/$(TOOL)
+	/tmp/$(TOOL) --logtostderr --v=6 -i ./output_tests/wholepkg -O zz_generated
+	@if ! diff -q ./output_tests/wholepkg/zz_generated.expected ./output_tests/wholepkg/zz_generated.go; then \
```
thockin: why is this different than https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/Makefile#L11 ?

examples/defaulter-gen/output_tests/wholepkg/a.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+type Stuct_With_Field_With_Field_Has_Manual_Defaulter struct {
+	TypeMeta struct{}
+	Stuct_With_Field_Has_Manual_Defaulter
+	OtherField string
+}
+
+// No defaulter will be generated
+type Stuct_No_Field_Has_Manual_Defaulter struct {
```
thockin: s/Stuct/Struct/g

examples/defaulter-gen/Makefile
caesarxuchao: The generated code depends on apimachinery so it does not compile and fails travis. So i changed the suffix to `.expected.`

examples/defaulter-gen/output_tests/wholepkg/a.go
caesarxuchao: Fixed.

examples/defaulter-gen/Makefile
lavalamp: Why does this exist? The library has a --verify-only option specifically so that we would not write things like this any more.

caesarxuchao: We can change the default-gen Makefile to use --verify-only in a follow-up PR.

For this PR, because the generated code does not compile and break travis, i need to change the suffix of the output file to .expected, so the --verify-only approach doesn't work.

To use --verify-only, we have two options:
* vendor a mock apimachinery folder in the test package to make the code compile.
* hack the .travis to ignore this package.

@lavalamp please let me know how you want me to proceed. Thanks.

thockin: blech wrt having external deps.  OK.  Can you comment that?

The point of checking it in is so that a human can look at a series of trivial output tests, and comprehend the output. --verify-only doesn't give me that?

lavalamp:
--verify-only tells you that the checked-in thing is still indeed what the
generator would output.

thockin:
In the deepcopy case, we regenerate and `git diff` which makes it easy for
people to produce the thing they need to check in.

In this case, it doesn't compile (we need to fix that, sigh...) so the same
trick doesn't work.  We need a simple way for anyone hacking to regen that
code for submit.

caesarxuchao:
How about split to 'make all' and 'make verify'?

I still need to bypass the compile issue, how about the "vendor a mock apimachinery package" trick?

mbohlool:
@thockin 

> In the deepcopy case, we regenerate and `git diff` which makes it easy for
> people to produce the thing they need to check in.

I disagree with this strategy for tests. Being able to check in a different test output to pass the test make it easier for people to bypass tests and don't look at the change before checking in. When people forced to fix test cases manually it more likely for them to spot a problem vs just run the test and check in the diff.

caesarxuchao:
@mbohlool i think it's fine as long as the difference of the generated code is shown in the PR and gets reviewed, so I agree with @thockin here.

@thockin How about split to 'make' to 'make all' and 'make verify', and use the '--verify-only' flag when run 'make verify'?

I still need to bypass the compile issue, how about the "vendor a mock apimachinery package" trick?

thockin:
I don't care much - I just want it to fail in an obvious way so humans can
examine the changes.

PR #56, opened 2017-05-15
Revert "Fix if loop related lint failures in the deepcopy-gen"

liggitt:
This reverts commit a438c28332fb46fff47d61b9d115275885c5c818 since it produces code that won't compile (see https://github.com/kubernetes/gengo/pull/46/files#r116396781) so master is usable (we need to bump as a prereq of kubernetes/kubernetes#45294)... can revisit this if we decide we want to jump through scoping or variable name hoops to avoid this lint error (I personally don't think it's worth it)

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -523,12 +523,11 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 				sw.Do("}\n", nil)
 				sw.Do("(*out)[key] = *newVal\n", nil)
 			} else {
-				sw.Do("newVal, err := c.DeepCopy(&val)\n", nil)
-				sw.Do("if err != nil {\n", nil)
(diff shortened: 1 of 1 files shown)


liggitt:
@smarterclayton PTAL
/cc @thockin @spzala

smarterclayton:
Yeah, the empty blocks were there for a reason.

/lgtm

spzala:
@liggitt thanks for fixing it Jordan. My bad, in my manual verification of generated code and build I apparently didn't run into it. I guess Gengo Travis passed because the method with nil return is not in Gengo but K8s. @thockin was quick in seeing a need for tests when lint related these changes were merged and kindly created new tests but those tests still missed few scenarios. I have also create a follow up PR (https://github.com/kubernetes/gengo/pull/51) to add new tests.  I value your opinion on spending time for lint may not worth :-) but if we can fix existing code on lint and going forward have lint complied code then that will reduce skipping lint check and produce more clean code. Thanks!!

PR #58, opened 2017-05-17
Update output from parse.go to improve use of the word error

cblecker:
The main reason for this stems from https://github.com/kubernetes/test-infra/issues/2783 and log highlighting in gubernator.

If we're ignoring type checking on a package, then if we can stay away from the word "error", this would massively cut down on the logs in the verify job.

Example verify with many type checker errors that end up being ignored: https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/ci-kubernetes-verify-master/4877/?log

cc: @brendandburns @fejta @rmmh


fejta:
/assign @brendandburns 

Note that I cannot merge in this repo

thockin:
I love this PR. Thanks.

PR #61, opened 2017-06-09
Fix defaulter-gen for recursive types

nikhita:
For recursive types, the defaulter currently goes into an infinite recursion.
Fix to support recursive types.

/cc @sttts

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -320,7 +320,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 				if d.object != nil {
 					continue
 				}
-				if buildCallTreeForType(t, true, existingDefaulters, newDefaulters) != nil {
+				if newCallTreeForType(existingDefaulters, newDefaulters).build(t, true) != nil {
 					args := defaultingArgsFromType(t)
 					sw.Do("$.inType|objectdefaultfn$", args)
 					newDefaulters[t] = defaults{
@@ -387,7 +387,22 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 	return packages
 }
 
-// buildCallTreeForType creates a tree of paths to fields (based on how they would be accessed in Go - pointer, elem,
+// callTreeForType contains fields necessary to build a tree for types.
+type callTreeForType struct {
+	existingDefaulters     defaulterFuncMap
+	newDefaulters          defaulterFuncMap
+	currentlyBuildingTypes map[*types.Type]bool
+}
+
+func newCallTreeForType(existingDefaulters, newDefaulters defaulterFuncMap) *callTreeForType {
+	return &callTreeForType{
+		existingDefaulters:     existingDefaulters,
+		newDefaulters:          newDefaulters,
+		currentlyBuildingTypes: make(map[*types.Type]bool),
+	}
+}
+
+// build creates a tree of paths to fields (based on how they would be accessed in Go - pointer, elem,
 // slice, or key) and the functions that should be invoked on each field. An in-order traversal of the resulting tree
 // can be used to generate a Go function that invokes each nested function on the appropriate type. The return
 // value may be nil if there are no functions to call on type or the type is a primitive (Defaulters can only be
@@ -396,7 +411,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 // that could be or will be generated. If newDefaulters has an entry for a type, but the 'object' field is nil,
 // this function skips adding that defaulter - this allows us to avoid generating object defaulter functions for
 // list types that call empty defaulters.
-func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap) *callNode {
+func (c *callTreeForType) build(t *types.Type, root bool) *callNode {
 	parent := &callNode{}
 
 	if root {
@@ -404,8 +419,8 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
 		parent.elem = true
 	}
 
-	defaults, _ := existingDefaulters[t]
-	newDefaults, generated := newDefaulters[t]
+	defaults, _ := c.existingDefaulters[t]
+	newDefaults, generated := c.newDefaulters[t]
 	switch {
 	case !root && generated && newDefaults.object != nil:
 		parent.call = append(parent.call, newDefaults.object)
@@ -432,19 +447,33 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
 	// base has been added already, now add any additional defaulters defined for this object
 	parent.call = append(parent.call, defaults.additional...)
 
+	// if the type already exists, don't build the tree for it and don't generate anything.
+	// This is used to avoid recursion for nested recursive types.
+	if c.currentlyBuildingTypes[t] {
+		return nil
+	}
+	// if type doesn't exist, mark it as existing
+	c.currentlyBuildingTypes[t] = true
+
+	defer func() {
+		// The type will now acts as a parent, not a nested recursive type.
+		// We can now build the tree for it safely.
+		c.currentlyBuildingTypes[t] = false
+	}()
+
 	switch t.Kind {
 	case types.Pointer:
-		if child := buildCallTreeForType(t.Elem, false, existingDefaulters, newDefaulters); child != nil {
+		if child := c.build(t.Elem, false); child != nil {
 			child.elem = true
 			parent.children = append(parent.children, *child)
 		}
 	case types.Slice, types.Array:
-		if child := buildCallTreeForType(t.Elem, false, existingDefaulters, newDefaulters); child != nil {
+		if child := c.build(t.Elem, false); child != nil {
 			child.index = true
 			parent.children = append(parent.children, *child)
 		}
 	case types.Map:
-		if child := buildCallTreeForType(t.Elem, false, existingDefaulters, newDefaulters); child != nil {
+		if child := c.build(t.Elem, false); child != nil {
 			child.key = true
 			parent.children = append(parent.children, *child)
 		}
@@ -458,13 +487,13 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
 					name = field.Type.Name.Name
 				}
 			}
-			if child := buildCallTreeForType(field.Type, false, existingDefaulters, newDefaulters); child != nil {
+			if child := c.build(field.Type, false); child != nil {
 				child.field = name
 				parent.children = append(parent.children, *child)
 			}
 		}
 	case types.Alias:
-		if child := buildCallTreeForType(t.Underlying, false, existingDefaulters, newDefaulters); child != nil {
+		if child := c.build(t.Underlying, false); child != nil {
 			parent.children = append(parent.children, *child)
 		}
 	}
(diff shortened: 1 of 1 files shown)


examples/defaulter-gen/generators/defaulter.go
```diff
@@ -320,7 +320,8 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 				if d.object != nil {
 					continue
 				}
-				if buildCallTreeForType(t, true, existingDefaulters, newDefaulters) != nil {
+				existingTypes := make(map[string]bool)
```
nikhita: PTAL if this is the right way to do it.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -396,7 +397,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 // that could be or will be generated. If newDefaulters has an entry for a type, but the 'object' field is nil,
 // this function skips adding that defaulter - this allows us to avoid generating object defaulter functions for
 // list types that call empty defaulters.
-func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap) *callNode {
+func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap, existingTypes map[string]bool) *callNode {
```
sttts: Can you document existingTypes? Is it feasible to have a recursive type in some test? Are there tests here in the generator?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -432,19 +433,26 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
 	// base has been added already, now add any additional defaulters defined for this object
 	parent.call = append(parent.call, defaults.additional...)
 
+	// if the type already exists, don't build the tree for that
+	if existingTypes[t.Name.Name] {
+		return parent
+	}
+	// if type doesn't exist, mark it as existing
+	existingTypes[t.Name.Name] = true
```
sttts: shouldn't we check that `Name` is actually files, i.e. not an anonymous type? Otherwise, we stop too early for those, don't we?

examples/defaulter-gen/generators/defaulter.go
nikhita: Will document it.
Currently there are no tests. But I'll add one.

examples/defaulter-gen/generators/defaulter.go
nikhita: So I should check for `t.Name`?

sttts: If I read the type godoc correctly, Name can be empty.

sttts:
Can you link a commit for the Kubernetes defaulters? Do they change after this PR?

thockin:
This could really use output tests like https://github.com/kubernetes/gengo/tree/master/examples/deepcopy-gen/output_tests

nikhita:
Updated the defaulter code. PTAL.

I have added some tests. However, I come across the following error when I run `make`: https://gist.github.com/nikhita/379c3789b8049038109b1ea1509b483b. It works when there is no `defaults.go` file though. I have also pushed the file generated in this case.

After making the changes in this PR and fixing the conversion in https://github.com/kubernetes/kubernetes/pull/47263, nothing breaks in Kubernetes (tested against HEAD). This PR shows the cumulative changes in both the defaulter and converter: https://github.com/nikhita/kubernetes/pull/3.

examples/defaulter-gen/output_tests/builtins/defaults.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
nikhita: Oops, will fix the year after the review.

nikhita:
The tests are being added in https://github.com/kubernetes/gengo/pull/55. Will create a PR to add a test for recursive type on top of https://github.com/kubernetes/gengo/pull/55.

nikhita:
Added tests on top of #55 here: https://github.com/caesarxuchao/gengo/pull/1.

sttts:
lgtm.

@thockin ptal at https://github.com/caesarxuchao/gengo/pull/1/files for example output.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -320,7 +320,10 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 				if d.object != nil {
 					continue
 				}
-				if buildCallTreeForType(t, true, existingDefaulters, newDefaulters) != nil {
+				// existingTypes keeps track of types that have already been visited in the tree.
+				// This is used to avoid recursion for recursive types.
+				existingTypes := make(map[*types.Type]bool)
```
sttts: s/existingTypes/alreadyVisitedTypes/

sttts:
@wojtek-t can this go in? (you are in the OWNERs files, together with @lavalamp. Though the merge+review policy here is unclear to me looking at previous PRs).

examples/defaulter-gen/generators/defaulter.go
nikhita: Done.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -458,20 +469,24 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
...
 		}
 	}
 	if len(parent.children) == 0 && len(parent.call) == 0 {
 		//glog.V(6).Infof("decided type %s needs no generation", t.Name)
 		return nil
 	}
+
+	// The type now acts as a parent, not a nested recursive type.
+	// We can now build the tree for it safely.
+	alreadyVisitedTypes[t] = false
```
lavalamp: The name is wrong if you set it to false after you are done. Should be "currentlyBuildingTypes" or "typesInProgress" or something.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -396,7 +399,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 // that could be or will be generated. If newDefaulters has an entry for a type, but the 'object' field is nil,
 // this function skips adding that defaulter - this allows us to avoid generating object defaulter functions for
 // list types that call empty defaulters.
-func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap) *callNode {
+func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap, alreadyVisitedTypes map[*types.Type]bool) *callNode {
```
lavalamp: Consider making a struct to hold all of the parameters that don't change, and make this function a method on that struct. Will make the code a lot easier to read.

lavalamp:
(happy to lgtm & merge after those comments are addressed)

nikhita:
@lavalamp Updated.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -424,32 +447,43 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
...
+		return nil
+	}
+	// if type doesn't exist, mark it as existing
+	c.currentlyBuildingTypes[c.t] = true
+
+	switch c.t.Kind {
 	case types.Pointer:
-		if child := buildCallTreeForType(t.Elem, false, existingDefaulters, newDefaulters); child != nil {
+		r := NewCallTreeForType(c.t.Elem, false, c.existingDefaulters, c.newDefaulters, c.currentlyBuildingTypes)
+		if child := r.buildCallTreeForType(); child != nil {
```
lavalamp: All of these should look like `if child := r.Build(c.t.Elem, false); child != nil {`

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -396,22 +419,22 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 // that could be or will be generated. If newDefaulters has an entry for a type, but the 'object' field is nil,
 // this function skips adding that defaulter - this allows us to avoid generating object defaulter functions for
 // list types that call empty defaulters.
-func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap) *callNode {
+func (c callTreeForType) buildCallTreeForType() *callNode {
```
lavalamp: `func (c *callTreeForType) Build(t *types.Type, root bool) *callNode {`

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -320,7 +320,11 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 				if d.object != nil {
 					continue
 				}
-				if buildCallTreeForType(t, true, existingDefaulters, newDefaulters) != nil {
+				// currentlyBuildingTypes keeps track of types that have already been visited in the tree.
+				// This is used to avoid recursion for recursive types.
+				currentlyBuildingTypes := make(map[*types.Type]bool)
+				c := NewCallTreeForType(t, true, existingDefaulters, newDefaulters, currentlyBuildingTypes)
+				if c.buildCallTreeForType() != nil {
```
lavalamp: This should be a one-liner:

```
if NewCallTreeForType(existingDefaulters, newDefaulters).Build(t, true) != nil {
```

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -387,6 +391,25 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
...
+// callTreeForType contains fields necessary to call buildCallTreeForType.
+type callTreeForType struct {
+	t                      *types.Type
+	root                   bool
+	existingDefaulters     defaulterFuncMap
+	newDefaulters          defaulterFuncMap
+	currentlyBuildingTypes map[*types.Type]bool
+}
+
+func NewCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefaulters defaulterFuncMap, currentlyBuildingTypes map[*types.Type]bool) callTreeForType {
```
lavalamp: Actually both this and Build should be private.

nikhita: We can't initialize currentlyBuildingTypes in newCallTreeForType because when this is encountered:

```
r := newCallTreeForType(c.existingDefaulters, c.newDefaulters)
	if child := r.build(t.Elem, false); child != nil
```

it will create a new map for currentlyBuildingTypes - destroying our purpose for storing the types that are currently building. :)

nikhita:
@lavalamp Updated. PTAL, thanks.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -432,19 +450,30 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
...
+	if c.currentlyBuildingTypes[t] {
+		return nil
+	}
+	// if type doesn't exist, mark it as existing
+	c.currentlyBuildingTypes[t] = true
+
 	switch t.Kind {
 	case types.Pointer:
-		if child := buildCallTreeForType(t.Elem, false, existingDefaulters, newDefaulters); child != nil {
+		r := newCallTreeForType(c.existingDefaulters, c.newDefaulters, c.currentlyBuildingTypes)
```
lavalamp: What I was trying to communicate is that you *don't* call newCallTreeForType here, you reuse `c`.

nikhita: Right, sorry - should have done this in the first place.

Updated. Please check if this is better.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -458,20 +481,24 @@ func buildCallTreeForType(t *types.Type, root bool, existingDefaulters, newDefau
...
 		}
 	}
 	if len(parent.children) == 0 && len(parent.call) == 0 {
 		//glog.V(6).Infof("decided type %s needs no generation", t.Name)
 		return nil
 	}
+
+	// The type now acts as a parent, not a nested recursive type.
+	// We can now build the tree for it safely.
+	c.currentlyBuildingTypes[t] = false
```
lavalamp: Looking good--last question, should this be in a defer block? The return on line 496 skips it, for example.

nikhita: Yes. Done.

lavalamp:
LGTM, thanks for bearing with me :)

PR #65, opened 2017-07-10
generator: include canonical import path when generating files

ericchiang:
Canonical import paths force `go {get,build,test,install}` to
respect the vanity import path of files. For client-go, this means
users will get an error if they place it in the wrong GOPATH path.
For example, `$GOPATH/src/github.com/kubernetes/client-go` would
fail. This import comment is ignored if the file is placed in a
"vendor" directory.

Have the generator include the import comment when generating
packages.

Ref: https://github.com/kubernetes/client-go/issues/223


thockin:
Can you rebuild tests in examples/deepcopy? That will regenerate example code.

ericchiang:
Updated

ericchiang:
cc @kubernetes/sig-api-machinery-pr-reviews

lavalamp:
Why would we change the import path in the generated files? There are plenty of non-generated files in the client to add a canonical import path to. It only takes one file with the import line to force the user to get it right.

I think generated files are a bad choice for this. Eventually we will want a build system that doesn't keep generated files around. Adding this will then have the result of occasionally making things not compile in a non-obvious way.

ericchiang:
@lavalamp good point. I've opened https://github.com/kubernetes/kubernetes/pull/48892 instead.

PR #67, opened 2017-07-25
deepcopy: remove legacy scheme registration

sttts:
This reverts commit 56b0b4977d7a7f68fad39721219afb6b29df7d8f.

With https://github.com/kubernetes/kubernetes/pull/53525 this is not necessary anymore.


sttts:
@lavalamp @thockin I want to go forward in kube and remove the unuser func registration for deepcopy. Can one of you merge this? Will revendor then immediately in kube.

thockin:
I admit I didn't beat up on the static DeepCopy stuff - how do we handle (or do we?) the case of embedded types having a DeepCopy() method, but the outer type needing one to be generated?

sttts:
> I admit I didn't beat up on the static DeepCopy stuff - how do we handle (or do we?) the case of embedded types having a DeepCopy() method, but the outer type needing one to be generated?

We don't need that luckily. It's the problem we had with ugorji and the topological order, but ugorji was based on reflection. I believe we can solve it on the gengo level as we are based on the Golang parser. It gives us the origin of a method, so we could detect that, maybe we even do out of pure implementation luck. Might be worth a test-case here in the deepcopy generator. But that's all unrelated to this PR :)

sttts:
@thockin we have that already and the generated code looks good: https://github.com/kubernetes/gengo/blob/a906434f2ded95d7376b3ec292d7c848d1cc5500/examples/deepcopy-gen/output_tests/wholepkg/a.go#L53

PR #69, opened 2017-07-28
Fix error for not matched names in comments

### LINKED ISSUE #68: Method name in comment does not match generated code by deepcopy-gen
[pkg/apis/apps/zz_generated.deepcopy.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/zz_generated.deepcopy.go#L88-L96):
```go
// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, creating a new ControllerRevision.
func (x *ControllerRevision) DeepCopy() *ControllerRevision {
	if x == nil {
		return nil
	}
	out := new(ControllerRevision)
	x.DeepCopyInto(out)
	return out
}
```

The method name in comment does not match the real method name.

supereagle:
Fix #68


k82cn:
@smarterclayton

PR #70, opened 2017-08-03
deepcopy: add +k8s:deepcopy-gen:skip-field tag and postDeepCopy func

sttts:
```
// If a "func (in *T) postDeepCopy(out *T)" func is predefined on an object by the
// developer, it is invoked after the generated deepcopy code. This can be used
// in combination with a +k8s:deepcopy-gen:skip-field tag above a field to skip
// deepcopy code generation for that field and to allow custom deepcopy code on
// a per-field basis.
```


sttts:
@thockin @wojtek-t ptal. This gives us a way out for types (like `interface{}`) the generator cannot cope with. We need this for CRD validation because it includes free form JSON in JSON schemas.

/cc @nikhita

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -470,15 +495,29 @@ func extractNonPointerInterfaces(comments []string) (bool, error) {
 	if len(values) == 0 {
 		return false, nil
 	}
-	result := values[0] == "true"
+	result := values[0] == "true" || values[0] == ""
```
thockin: We do a poor job overall of normalizing and validating these tags.  Would be nice to not have to consider "" everywhere.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -470,15 +495,29 @@ func extractNonPointerInterfaces(comments []string) (bool, error) {
...
 	return result, nil
 }
 
+func extractSkipField(comments []string) (bool, error) {
+	values := types.ExtractCommentTags("+", comments)[skipFieldTagName]
+	if len(values) == 0 {
+		return false, nil
+	}
+	result := values[0] == "true" || values[0] == ""
+	for _, v := range values {
```
thockin: can't we just assert len() == 0 ?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -642,7 +681,7 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 			sw.Do("for key := range *in {\n", nil)
 			sw.Do("(*out)[key] = struct{}{}\n", nil)
 			sw.Do("}\n", nil)
-		case t.Elem.IsAssignable():
+		case !hasPostDeepCopyMethod(t.Elem) && t.Elem.IsAssignable():
```
thockin: Does this mean the existence of the (private) method inhibits copying?  Why do we need the tag, then?  Shouldn't those be orthogonal concepts?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -719,6 +758,8 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
 			sw.Do("}\n", nil)
 		} else if t.Elem.Kind == types.Struct {
 			sw.Do("(*in)[i].DeepCopyInto(&(*out)[i])\n", nil)
+		} else if t.Elem.Kind == types.Alias {
```
thockin: a) this should just be an ||
b) if it is an Alias, you still need to check the underlying Kind, no?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -815,8 +870,20 @@ func (g *genDeepCopy) doPointer(t *types.Type, sw *generator.SnippetWriter) {
...
 
 func (g *genDeepCopy) doAlias(t *types.Type, sw *generator.SnippetWriter) {
-	// TODO: Add support for aliases.
-	g.doUnknown(t, sw)
+	unsafePointer := g.universe.Type(types.Name{Package: "unsafe", Name: "Pointer"})
+	g.imports.AddType(unsafePointer)
+
+	args := generator.Args{
+		"type":          t,
+		"underlaying":   t.Underlying,
```
thockin: s/laying/lying/

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -328,7 +352,7 @@ func copyableType(t *types.Type) bool {
 		return false
 	}
 	// TODO: Consider generating functions for other kinds too.
-	if t.Kind != types.Struct {
+	if t.Kind != types.Struct && !(t.Kind == types.Alias && hasPostDeepCopyMethod(t)) {
```
sttts: This makes alias+hasPostDeepCopyMethod copyable. So the question is why the non-existence of postDeepCopy on an alias makes it uncopyable. I think it shouldn't.

examples/deepcopy-gen/generators/deepcopy.go
sttts: Our client-gen recently moved to accept `""` as true. IMO this is what one expects. Let me factor that comparison out. Then it's more explicit.

examples/deepcopy-gen/generators/deepcopy.go
sttts: You mean to drop the `=true` variant? I am fine with that as well as long as it is consistent.

You mean the following?
```
tag => true
tag=true => invalid value
tag=false => false
```

examples/deepcopy-gen/generators/deepcopy.go
sttts: Yes, I mix two orthogonal concept in this commit: postDeepCopy() and `skip-field` tag. Would make sense to split them apart in two commits indeed.

The reason for this line is: I want `postDeepCopy` to be private. If we use assignment and `val` has a `postDeepCopy` here, we have to call `postDeepCopy` from `t`'s `DeepCopyInto` func. If `t` and `val` are from different packages, this won't work. So we switch to  the mode where don't assign, but call `DeepCopy` on `val` (then the `postDeepCopy` is invoked from `val`'s `DeepCopyInto` func).

I will add a `PkgPath` comparison.

sttts: a pity, `Type` has no `PkgPath`.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -309,6 +312,27 @@ func hasDeepCopyMethod(t *types.Type) bool {
...
+// finished.
+//
+//  The correct signature for a type T is:
+//    func (in *T) postDeepCopy(out *T)
+func hasPostDeepCopyMethod(t *types.Type) bool {
+	for mn, mt := range t.Methods {
+		if mn != "postDeepCopy" {
+			continue
+		}
+		if len(mt.Signature.Results) != 0 || len(mt.Signature.Parameters) != 1 {
```
sttts: Fixed.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -309,6 +312,27 @@ func hasDeepCopyMethod(t *types.Type) bool {
 	return false
 }
 
+// hasDeepCopyMethod returns true if an appropriate postDeepCopy(out *T) method is
```
sttts: Fixed.

examples/deepcopy-gen/generators/deepcopy.go
sttts: Factored out the alias change and moved it to https://github.com/kubernetes/gengo/pull/73. That's independent and nice to have.

examples/deepcopy-gen/generators/deepcopy.go
sttts: Fixed.

examples/deepcopy-gen/generators/deepcopy.go
sttts: factored out. done.

sttts:
@thockin ptal. Without the alias support this looks much better.

sttts:
@thockin friendly ping. Ptal.

PR #73, opened 2017-08-07
deepcopy: add alias support

sttts:
~~TODO:~~
- [x] verify that deepcopies in Kubernetes are correct.
- [x] verify that deepcopies in Openshift are correct.


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -762,6 +764,8 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
 			sw.Do("}\n", nil)
 		} else if t.Elem.Kind == types.Struct {
 			sw.Do("(*in)[i].DeepCopyInto(&(*out)[i])\n", nil)
+		} else if t.Elem.Kind == types.Alias {
```
sttts: > thockin:
> a) this should just be an ||
> b) if it is an Alias, you still need to check the underlying Kind, no?

sttts: @thockin I guess you are right. On the other hand we also fall back to `.DeepCopy` one case below. My change does not change much here, only optimized a bit. Struct aliases will work with this. For the full solution we need another case. Or fall back to `g.generateFor(t, sw)` with another `{ in, out := ... }` scope.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -309,6 +309,30 @@ func hasDeepCopyMethod(t *types.Type) bool {
 	return false
 }
 
+// hasDeepCopyMethod returns true if an appropriate postDeepCopy(out *T) method is
```
thockin: comment doesn't match function name

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -309,6 +309,30 @@ func hasDeepCopyMethod(t *types.Type) bool {
...
+// defined for the given type. It is called after the generated deepcopy is
+// finished.
+//
+//  The correct signature for a type T is:
+//    func (in *T) postDeepCopy(out *T)
```
thockin: it feels dirty to be looking at private methods.  Should this be public?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -770,7 +796,7 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 		case types.Struct:
 			if hasMethod {
 				sw.Do("out.$.name$ = in.$.name$.DeepCopy()\n", args)
-			} else if t.IsAssignable() {
+			} else if !hasPostDeepCopyMethod && t.IsAssignable() {
```
thockin: why does this care about the method?  Please comment these - it's really twisty

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -471,6 +475,10 @@ func (g *genDeepCopy) needsGeneration(t *types.Type) bool {
 	return true
 }
 
+func tagValueIsTrue(val string) bool {
```
thockin: This seems like something that will surprise me later.  Can you at least comment why "" is considered true here?  Better, rename it to be clear

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -494,15 +502,29 @@ func extractNonPointerInterfaces(comments []string) (bool, error) {
...
 }
 
+func extractSkipField(comments []string) (bool, error) {
+	values := types.ExtractCommentTags("+", comments)[skipFieldTagName]
+	if len(values) == 0 {
+		return false, nil
+	}
+	result := tagValueIsTrue(values[0])
+	for _, v := range values {
+		if tagValueIsTrue(v) && values[0] != "" {
```
thockin: I don't understand why this is comparing to [0] ?

examples/deepcopy-gen/main.go
```diff
@@ -22,7 +22,10 @@ limitations under the License.
 // generating new ones.
 //
 // If a "func (in *T) postDeepCopy(out *T)" func is predefined on an object by the
```
thockin: Why can't this logic just check for and call the T.DeepCopy()?  If the use case is to not generate code and then call custom code, that is exactly what defining DeepCopy does.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -356,7 +357,7 @@ func copyableType(t *types.Type) bool {
 		return false
 	}
 	// TODO: Consider generating functions for other kinds too.
-	if t.Kind != types.Struct {
+	if t.Kind != types.Struct && t.Kind != types.Alias {
```
thockin: I really need you to explain what this is doing and why?  Now we deepcopy other types beside struct?  I'm OK with that but why Alias?

sttts:
> Remind me what the need for this is? I'm liking it less and less as I think about it and the precedent it sets.

Mostly for completeness actually, as you can define funcs on aliases (DeepCopy or postDeepCopy). We don't require it anyway. That's why I moved it out. Am not a fan of the complexity either.

sttts:
@thockin "just for completeness" turned into a real use-case, or even a number of them in the OpenShift types. I rebased and fixed a couple of issues.

soltysh:
I've double checked it, since I hit this while working on 1.8.1 rebase into OpenShift. This fix is properly addressing aliases in deep copies generation.

examples/deepcopy-gen/generators/deepcopy.go
smarterclayton: You should probably be doing the lookup here to ensure the alias is to a struct.

examples/deepcopy-gen/generators/deepcopy.go
smarterclayton: If you get an alias, the alias needs to be to a struct.

deads2k:
@sttts rebase please

sttts:
> rebase please

done

sttts:
@soltysh ptal. In every of the `do*` funcs the given type can be an alias. The same is true for the value of a slice or map. We always pass down the alias type, but do the actual copy code according to the underlying type. There are no changes in Kube with this change.

deads2k:
test failures

sttts:
> test failures

They look infra related.

sttts:
@soltysh doublechecked that origin deepcopies look good. They do!

sttts:
Tests are not green because of https://github.com/kubernetes/gengo/pull/91

examples/deepcopy-gen/output_tests/aliases/doc.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
soltysh: Should probably be 2018 by now ;)

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -738,34 +770,32 @@ func (g *genDeepCopy) doInterface(t *types.Type, sw *generator.SnippetWriter) {
 }
 
 func (g *genDeepCopy) doPointer(t *types.Type, sw *generator.SnippetWriter) {
+	ut := underlyingType(t)
+	uet := underlyingType(ut.Elem)
+
 	sw.Do("if *in == nil { *out = nil } else {\n", t)
-	if hasDeepCopyMethod(t.Elem) {
-		sw.Do("*out = new($.Elem|raw$)\n", t)
+	if hasDeepCopyMethod(ut.Elem) {
```
soltysh: I think it's worth the comment you left me [here](https://github.com/kubernetes/gengo/pull/73#issuecomment-354788061) why we check `ut.Elem` but generate for `uet`. It's something that needs to be cleared to everyone reading the code.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -578,93 +618,96 @@ func (g *genDeepCopy) doBuiltin(t *types.Type, sw *generator.SnippetWriter) {
...
 			sw.Do("for key := range *in {\n", nil)
 			sw.Do("(*out)[key] = struct{}{}\n", nil)
 			sw.Do("}\n", nil)
-		case t.Elem.IsAssignable():
+		case uet.IsAssignable():
 			sw.Do("for key, val := range *in {\n", nil)
 			sw.Do("(*out)[key] = val\n", nil)
 			sw.Do("}\n", nil)
-		case t.Elem.Kind == types.Interface:
+		case ut.Elem.Kind == types.Interface: // only call this for non-aliased interfaces
```
soltysh: If `ut.Elem` is an interface then `underlyingType` won't change it. Why not using `uet.Kind` here in that case?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -578,93 +618,96 @@ func (g *genDeepCopy) doBuiltin(t *types.Type, sw *generator.SnippetWriter) {
...
+			g.generateFor(ut.Elem, sw)
 			sw.Do("}\n", nil)
-		} else if hasDeepCopyMethod(t.Elem) {
+		} else if hasDeepCopyMethod(ut.Elem) {
 			sw.Do("(*out)[i] = (*in)[i].DeepCopy()\n", nil)
 			// REVISIT(sttts): the following is removed in master
 			//} else if t.Elem.IsAssignable() {
 			//	sw.Do("(*out)[i] = (*in)[i]\n", nil)
-		} else if t.Elem.Kind == types.Interface {
+		} else if ut.Elem.Kind == types.Interface {
```
soltysh: Same here, wrt `uet.Kind` instead of `ut.Elem.Kind`?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -683,46 +728,33 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
...
+			g.generateFor(t, sw)
+			sw.Do("}\n", nil)
+		case uft.Kind == types.Struct:
+			if t.IsAssignable() {
 				sw.Do("out.$.name$ = in.$.name$\n", args)
 			} else {
 				sw.Do("in.$.name$.DeepCopyInto(&out.$.name$)\n", args)
 			}
-		case types.Interface:
+		case t.Kind == types.Interface: // only call this for non-aliased interfaces
```
soltysh: `uft.Kind` ?

examples/deepcopy-gen/output_tests/aliases/doc.go
sttts: We don't update unchanged files, do we?

examples/deepcopy-gen/generators/deepcopy.go
sttts: Actually this does not matter. The golang parser renders interface aliases as interfaces, without access to the underlying type. I fixed this, added comments and examples.

examples/deepcopy-gen/generators/deepcopy.go
sttts: dito

examples/deepcopy-gen/generators/deepcopy.go
sttts: dito

examples/deepcopy-gen/generators/deepcopy.go
sttts: done

sttts:
@soltysh addressed your comments. ptal.

examples/deepcopy-gen/output_tests/aliases/doc.go
sttts: fixed

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -561,110 +603,127 @@ func (g *genDeepCopy) generateFor(t *types.Type, sw *generator.SnippetWriter) {
...
 		default:
 			sw.Do("for key, val := range *in {\n", nil)
-			if g.copyableAndInBounds(t.Elem) {
-				sw.Do("newVal := new($.|raw$)\n", t.Elem)
+			if g.copyableAndInBounds(uet) {
+				sw.Do("newVal := new($.|raw$)\n", ut.Elem)
 				sw.Do("val.DeepCopyInto(newVal)\n", nil)
 				sw.Do("(*out)[key] = *newVal\n", nil)
-			} else if t.Elem.Kind == types.Slice && t.Elem.Elem.Kind == types.Builtin {
+			} else if uet.Kind == types.Slice || uet.Kind == types.Builtin {
```
soltysh: Should be `&&`.

sttts: fixed

sttts:
Test fail because of https://github.com/kubernetes/gengo/pull/91.

examples/deepcopy-gen/output_tests/wholepkg/zz_generated.go
```diff
@@ -761,9 +763,7 @@ func (in *Struct_Slices_Alias) DeepCopyInto(out *Struct_Slices_Alias) {
 			(*out)[i] = (*in)[i].DeepCopy()
 		}
 	}
-	if in.ManualSliceField != nil {
```
thockin: These deltas are suspicious...  Is it right?  Why?

thockin:
Can you update the first comment in the PR (the "main" comment) to describe what this is doing and why?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -292,21 +292,40 @@ func (g *genDeepCopy) copyableAndInBounds(t *types.Type) bool {
...
+	if len(f.Signature.Parameters) != 0 {
+		return false
+	}
+	if len(f.Signature.Results) != 1 || f.Signature.Results[0].Name != t.Name {
+		return false
+	}
+	return true
+}
+
+// hasDeepCopyIntoMethod returns true if an appropriate DeepCopyInit() method is
```
thockin: Init() or Into() ?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -292,21 +292,40 @@ func (g *genDeepCopy) copyableAndInBounds(t *types.Type) bool {
...
+//    func (t *T) DeepCopyInit(t *T)
+func hasDeepCopyIntoMethod(t *types.Type) bool {
+	f, found := t.Methods["DeepCopyInto"]
+	if !found {
+		return false
+	}
+	if len(f.Signature.Parameters) != 1 {
+		return false
+	}
+	if len(f.Signature.Results) != 0 {
```
thockin: shouldn't this also be checking the type of the argument (kind == pointer, etc)?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -327,17 +346,38 @@ func copyableType(t *types.Type) bool {
...
 	}
-	// Also, filter out private types.
-	if namer.IsPrivateGoName(t.Name.Name) {
+
+	if t.Kind == types.Alias {
+		// if the underlying built-in is not deepcopy-able, deepcopy is opt-in through definition of custom methods.
+		if hasDeepCopyMethod(t) || hasDeepCopyIntoMethod(t) {
+			return true
+		} else {
+			return t.Underlying.Kind != types.Builtin || copyableType(t.Underlying)
```
thockin: won't this catch Map and Slice when it didn't used to?

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -327,17 +346,38 @@ func copyableType(t *types.Type) bool {
...
+
+	if t.Kind != types.Struct {
 		return false
 	}
+
 	return true
 }
 
+func underlyingType(t *types.Type) *types.Type {
+	for {
```
thockin:
```
for t.Kind == types.Alias {
    t = t.Underlying
}
```
 
?

examples/deepcopy-gen/output_tests/aliases/doc.go
```diff
@@ -0,0 +1,89 @@
...
+	X int
+}
+
+type Interface interface {
+	DeepCopyInterface() Interface
+	DeepCopyAliasInterface() AliasInterface
+	DeepCopyAliasAliasInterface() AliasAliasInterface
+}
+
+type Builtin int
```
thockin: It would actually be way nicer to review / maintain a couple aliases added to each of the existing output_tests, rather than this.

e.g. aliases for int/string in output_tests/builtins, etc

thockin:
Does this produce deltas against kubernetes?

examples/deepcopy-gen/generators/deepcopy.go
sttts: Into()

examples/deepcopy-gen/generators/deepcopy.go
sttts: Actually, I tend to remove all these checks. IMO it's better to create code that is not type-correct than skipping a DeepCopy functions that the user expects to be called, but it isn't because there is slight type-error. If we don't want to rely on the type checker, we could also error out here in the generator. But skipping slightly off funcs leads to hard to understand wrong behaviour of the generated code.

examples/deepcopy-gen/generators/deepcopy.go
sttts: Yes, it does. It's intentional, compare https://goplay.space/#KWig91IjuQv

Will add a comment.

examples/deepcopy-gen/generators/deepcopy.go
sttts: :+1:

examples/deepcopy-gen/output_tests/aliases/doc.go
sttts: I think both ways have advantages, depending on your view point. To be complete with alias support I pretty much prefer this structure. We can maintain both if you feel strongly about this.

examples/deepcopy-gen/generators/deepcopy.go
sttts: I added another commit with fatal error. I would very much prefer that. If this turns out to be a problem on any external type with DeepCopy/Into, we can still add an opt-out tag. But we haven't seen such a case yet, so let's keep it simple for now.

sttts:
> Does this produce deltas against kubernetes?

Noticed an issue with aliased pointers, maps and slices and pre-defined DeepCopy funcs. On it....

sttts:
@thockin here is the diff of kube. Looked through most of the changes and they look. But more eyes on the changes would be good. https://gist.github.com/sttts/7474ed37382b7abc79eeaeb99a9929b6

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -286,27 +286,76 @@ func (g *genDeepCopy) copyableAndInBounds(t *types.Type) bool {
...
+	return f.Signature
+}
+
+// deepCopyIntoMethod returns the signature of a DeepCopyInto() method, nil or calls glog.Fatalf
+// if the type is wrong. DeepCopyInto allows more efficient deep copy
+// implementations to be defined by the type's author.  The correct signature
+// for a type T is:
+//    func (t T) DeepCopyInto(t *T)
+// or:
+//    func (t *T) DeepCopyInit(t *T)
```
jennybuckley: Won't this return nil for "func (t *T) DeepCopyInit(t *T)" ?

sttts: typo. fixed.

soltysh:
From notable changed the following types got `DeepCopyInto` methods:

* - `ExtraValue`
* - `ResourceList`
* - `SortableRuleSlice`
* - `ConfigurationMap`
* - `FailedNodesMap`
* - `HostPriorityList`
* - `JSONSchemaDefinitions`
* - `JSONSchemaDependencies`
* - `Verbs`
* - `ByGroupPriorityMinimum`

The few remaining changes look sane.

sttts:
> ExtraValue

This was actually wrong before, turning `nil` into a zero element slice.

> ResourceList

It's an alias and the deepcopy funcs are correct.

> SortableRuleSlice

It's an alias and the deepcopy funcs are correct.

> ConfigurationMap

It's an `map[string]string` alias and the deepcopy funcs are correct.

> FailedNodesMap

It's an `map[string]string` alias and the deepcopy funcs are correct.

> HostPriorityList

It's a struct slice alias with copyable structs, and the deepcopy funcs are correct.

> JSONSchemaDefinitions

It's a non-copyable JSONSchemaProps string map alias, and the deepcopy funcs are correct.

> JSONSchemaDependencies

It's a non-copyable JSONSchemaPropsOrStringArray string map alias, and the deepcopy funcs are correct.

> Verbs

It's a string slice alias, and the deepcopy funcs are correct.

> ByGroupPriorityMinimum

It's a non-copyable `*APIService` slice alias and the deepcopy funcs are correct.

sttts:
In short: we get a couple of alias `DeepCopy` funcs. But's that's intended.

soltysh:
SGTM, ship it!

sttts:
Rebased.

thockin:
PR comment says "TODO".  Please update.

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -286,27 +286,76 @@ func (g *genDeepCopy) copyableAndInBounds(t *types.Type) bool {
...
-		return true
+//    func (t *T) DeepCopy() *T
+func deepCopyMethod(t *types.Type) *types.Signature {
+	f, found := t.Methods["DeepCopy"]
+	if !found {
+		return nil
 	}
-	return false
+	if len(f.Signature.Parameters) != 0 {
+		glog.Fatalf("Type %v: invalid DeepCopy signature, expected no parameters", t)
```
thockin: Why are we aborting?  Other tools just ignore signatures that don't match.  e.g.

https://play.golang.org/p/tl05gzSswY-

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -286,27 +286,76 @@ func (g *genDeepCopy) copyableAndInBounds(t *types.Type) bool {
...
+}
+
+// deepCopyIntoMethod returns the signature of a DeepCopyInto() method, nil or calls glog.Fatalf
+// if the type is wrong. DeepCopyInto allows more efficient deep copy
+// implementations to be defined by the type's author.  The correct signature
+// for a type T is:
+//    func (t T) DeepCopyInto(t *T)
+// or:
+//    func (t *T) DeepCopyInto(t *T)
+func deepCopyIntoMethod(t *types.Type) *types.Signature {
```
thockin: should we perhaps demand that a user defines both of these or neither of these?  Or are we ok to mix-and-match?

thockin:
Travis fail:  examples/deepcopy-gen/generators/deepcopy_test.go:263: undefined: hasDeepCopyMethod

examples/deepcopy-gen/generators/deepcopy.go
sttts: Having a wrong signature is most probably a programmer error. And assuming the custom deepcopy being called, but isn't, sound dangerous and hard to debug. I prefer to fail early here.

examples/deepcopy-gen/generators/deepcopy.go
sttts: We are ok. One alone works for us. The DeepCopyInto is usually an optimization with less allocations.

examples/deepcopy-gen/generators/deepcopy.go
soltysh: I second that opinion. Chasing bugs like that, which can be easily caught earlier is definitely better option.

sttts:
> Travis fail: examples/deepcopy-gen/generators/deepcopy_test.go:263: undefined: hasDeepCopyMethod

Fixed tests and completed them with new cases and deepCopyInto tests.

sttts:
Thanks for merging. Updating the kube PR to revendor it.

PR #74, opened 2017-08-07
Split UT and Examples in different stage in travis

asifdxtreme:
Split UT and Examples in different stages in  Travis CI so that user can debug easily what went wrong in CI
Preview
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,16 @@
 language: go
 go:
-  - 1.6
-  - 1.7
-  - tip
-
+  - 1.8
+  
 go_import_path: k8s.io/gengo
 
-script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,13 +1,16 @@
 language: go
 go:
-  - 1.6
-  - 1.7
-  - tip
-
+  - 1.8
```
thockin: running against tip is interesting, too, no?

asifdxtreme: Added tip as per suggestion

asifdxtreme:
@thockin Thanks for your suggestion, I have added tip as per your suggestion

PR #84, opened 2017-11-27
Format files using goimports

allenpetersen:
Format files using x/tools/imports to get the import sort order from goimports.

This is to address https://github.com/kubernetes/kubernetes/issues/55542

### DIFF
--- a/generator/execute.go
+++ b/generator/execute.go
@@ -19,13 +19,13 @@ package generator
 import (
 	"bytes"
 	"fmt"
-	"go/format"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


allenpetersen:
I've signed the CLA.

thockin:
Did you run this against the codebase to see what changes it would make to examples?  Did you run it against kubernetes?

allenpetersen:
I did a kubernetes build with a local version of gengo and the generated files now look like they have been processed by goimports. It fixes the diff from the k8s issue.

I didn't see any dependency management with gengo but there will need to be a change to the kubernetes Godep to bring in this version as it now depends on "golang.org/x/tools/imports"

PR #86, opened 2017-11-27
Fix go vet warning

thockin:


thockin:
self-merge to fix travis

PR #90, opened 2017-12-28
Fix default-gen for []* type

fisherxu:
Now type slice are almost all non-pointer, like **Items []Pod**. 
But this([Items []*PartialObjectMetadata](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1alpha1/types.go#L160)) slice's elem/kind is ptr
And default-gen will hard-code add **&** before Item like a := **&in**.Spec.Template.Spec.Volumes[i] .
when Item's kind is ptr, it shouldn't add &. Otherwise it will generate **.

In my env, I add a private element in [ObjectMeta](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1alpha1/types.go#L151), and add manual default. So the generate_default will involve it and failed when compile.
image

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -470,6 +470,9 @@ func (c *callTreeForType) build(t *types.Type, root bool) *callNode {
 	case types.Slice, types.Array:
 		if child := c.build(t.Elem, false); child != nil {
 			child.index = true
+			if t.Elem.Kind == types.Pointer {
+				child.elem = true
+			}
 			parent.children = append(parent.children, *child)
 		}
 	case types.Map:
@@ -756,7 +759,12 @@ func (n *callNode) WriteMethod(varName string, depth int, ancestors []*callNode,
 	switch {
 	case n.index:
 		sw.Do("for $.index$ := range $.var$ {\n", vars)
-		sw.Do("$.local$ := &$.var$[$.index$]\n", vars)
+		if n.elem {
+			sw.Do("$.local$ := $.var$[$.index$]\n", vars)
+		} else {
+			sw.Do("$.local$ := &$.var$[$.index$]\n", vars)
+		}
+
 		n.writeCalls(local, true, sw)
 		for i := range n.children {
 			n.children[i].WriteMethod(local, depth+1, append(ancestors, n), sw)


sttts:
Do you  have a counterpart PR for kubernetes? Just to see that this does not break anything, i.e. no defaulters change there. Would be nice to have output tests as for the deepcopy example.

fisherxu:
@sttts Have no PR for kubernetes. I add a private field in v1.ObjectMeta and manual set default to it. So the    defaulters will change and generated codes.

fisherxu:
I think if other slice fields also become pointer, it will have same error.

sttts:
@fisherxu please just revendor gengo in kube and push the result of running the generators. I expect nothing to be changed, but just to be sure...

Also adding output tests as in the deepcopy would be awesome here.

fisherxu:
@sttts Thanks for reply :) I find that in deepcopy-gen, it has made a judgment about whether the slice'elem'kind  is a pointer, [the code-link](https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/generators/deepcopy.go#L662) .
So deepcopy should be the right way...

fisherxu:
And I also find the deepcopy codes related to **slice'elem'kind is a pointer**, only appears in the apigroup/file which I add the private field and it's right.

```
// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
func (in *PartialObjectMetadataList) DeepCopyInto(out *PartialObjectMetadataList) {
	*out = *in
	out.TypeMeta = in.TypeMeta
	if in.Items != nil {
		in, out := &in.Items, &out.Items
		*out = make([]*PartialObjectMetadata, len(*in))
		for i := range *in {
			if (*in)[i] == nil {
				(*out)[i] = nil
			} else {
				(*out)[i] = new(PartialObjectMetadata)
				(*in)[i].DeepCopyInto((*out)[i])
			}
		}
	}
	return
}
```
Now we have no defaulter to []*xxx, if added later, default-gen must be fixed ;)

sttts:
@fisherxu thanks for looking into deepcopy :) I was talking about copying the idea of the [output_tests directory](https://github.com/kubernetes/gengo/tree/master/examples/deepcopy-gen/output_tests) and the corresponding [Makefile rules](https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/Makefile), adapted to defaulter-gen. It's always hard to judge whether a generator change is correct if there are no examples. We don't want to break k8s.io/gengo by unexpected changes that would show up after we revendor into k8s.io/kubernetes.

fisherxu:
@sttts   yeah sure, I can try to add the test as deepcopy first :)

thockin:
We should have output_tests for defaulter-gen, too.

fisherxu:
@thockin @sttts Have picked this PR to defaulter-gen test [PR,](https://github.com/kubernetes/gengo/pull/93) where we can review the fix according to test code. PTAL:)

Issue #24, opened 2017-01-02
Copyright boilerplate changes YEAR, but shouldn't


thockin:
The copyrighted code is either copyright the first year it was generated or else it is copyright the year we wrote gengo.  It obviously doesn't change copyright by nature of being regenerated.

thockin:
Legal guidance is that the year is literally meaningless on these files.  My proposal is that we simply omit the year field, and fix the kubernetes linter(s) to accept boilerplate with or without year.

sttts:
> My proposal is that we simply omit the year field,

:+1:

PR #91, opened 2018-01-02
fix deepcopy/sets generate of new year and remove YEAR field

fisherxu:
fix the new year of 2018 and remove YEAR field
 fix issue:[#24](https://github.com/kubernetes/gengo/issues/24)


deads2k:
@sttts pick to openshift fork please.

sttts:
@deads2k we don't have boilerplate headers with YEARs.

lavalamp:
This is a bit wrong-- *new* files should get the new year, but *old* files should not change.

Additionally, this appears to just modify the generated output instead of fixing the generator?

sttts:
> This is a bit wrong-- new files should get the new year, but old files should not change.

I don't disagree. But is this a regression? I would prefer to get CI fixed first and do a proper fix in a follow-up. A quick look at the generator suggests that the code is pretty dumb just replacing the YEAR. We would have to write to a temporary file and do a fuzzy comparison.

sttts:
Doesn't look like a regression. We had the same in 2017: https://github.com/kubernetes/gengo/issues/24

thockin:
Agree - this sucks, but it's not on anyone's plate right now, and is not a trivial fix.

/lgtm

thockin:
Need to also regenerate sets ?

sttts:
Travis thinks so. @fisherxu can you update sets as well?

fisherxu:
@sttts @thockin Have updated sets in another PR, xref: [92](https://github.com/kubernetes/gengo/pull/92) , and also add .gitignore for deepcopy-gen, PTAL. :)

sttts:
@fisherxu can you merge both PR so that CI is happy? One alone won't be green ;)

fisherxu:
@sttts  sure, have done :)

sttts:
Lgtm. Maybe squash the two generation commits.

fisherxu:
@sttts Have squashed :)
 Also add .gitignore for deepcopy and Makefile for set-gen. @thockin PTAL.
and will add test for default-gen later in another PR.

fisherxu:
@thockin deepcopy and sets are all in this PR, can you help merge it? :)

fisherxu:
@thockin @sttts Gentle ping, please review this PR and we should make CI for other PR green.

fisherxu:
@thockin @lavalamp @sttts have removed the YEAR field in boilerplate, and regenerated, PTAL.

fisherxu:
If we delete this line [#105](https://github.com/kubernetes/gengo/blob/master/args/args.go#L105), all the generated files in k/k will also be regenerated.

sttts:
@fisherxu can you squash the first and last commit?

fisherxu:
@sttts Sure, have done.

sttts:
/lgtm

@thockin ptal.

thockin:
Did we fix the kubernetes/kubernetes checker to not fail on the lack of a year?

fisherxu:
@thockin This PR only affect generated files in gengo, kubernetes/kubernetes use another boilerplate in hack/boilerplate, and haven't fix it. Do we need to change generated files in kubernetes/kubernetes?

sttts:
> Do we also need to change generated files in kubernetes/kubernetes?If we need fix that

Yes, basically the same as you have done here. Plus we have a CI checker for boilerplate in place that needs to be updated. That checker will have to check for date for non-generated files, but allow a missing date for generated files. So it's more complicated. But it is a follow-up of this PR.

@thockin can we merge this to make progress? The follow-up (revendoring) will have to fix the things in kube.

sttts:
🎉

fisherxu:
@thockin @sttts I find all generated files can be falled into two categories, one class of filenames contains generate and the other is under fixed folder like pkg/client/lister. Can we list all the folders and all the files contains generate, then do no YEAR field check. Other files do like before?

fisherxu:
We can set generated_dirs like [skipped_dirs,](https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.py#L132) then do no YEAR field check, look forward to hearing from your opinions :)

sttts:
Yes, something in this direction. Haven't look at the boilerplate test script.

Note: there might be some files in those generated packages that are not generated (*_expansion.go). Not sure whether we actually use this feature in Kube though, maybe only in OpenShift. Just delete the folders, regenerate and compare which files are missing. Maybe that set is empty.

fisherxu:
Have sent a PR to remove all YEAR field in k/k and also fix the checker, cc @thockin @lavalamp @sttts , ptal :)

sttts:
@fisherxu I fear this conflicts with https://github.com/kubernetes/kubernetes/pull/57656

fisherxu:
@sttts yes, I have seen at [kubernetes/kubernetes#59172](https://github.com/kubernetes/kubernetes/pull/59172) , we can wait kubernetes/kubernetes#57656 merge first.

sttts:
@fisherxu working with @nikhita on it right now to get it in asap.

fisherxu:
@sttts  No problem, I can help take a look for the failed CItests :)

sttts:
@fisherxu fixing the path calculation right now. We don't want to hard-coded kube packages there.

fisherxu:
@sttts Ok, talking with @nikhita now :)

PR #94, opened 2018-01-11
Output 'Code generated by' comment if arg is set

jennybuckley:
Upstreaming kubernetes/kubernetes#58115

The format of the comment is:
```
// Code generated by <gemerator-name>. DO NOT EDIT.
```

It will be located directly above the package declaration.


sttts:
Do we need a way to customize the string? If the output language is not Go (do we have that? openapi or protobuf?), the string must be customizable (not via flag, but via code, i.e. optionally).

lavalamp:
Yes, one of these generators makes protobuf...

jennybuckley:
I believe the proto files already have "// This file was autogenerated by go-to-protobuf. Do not edit it manually!" which this pr shouldn't affect.

args/args.go
```diff
@@ -92,6 +96,7 @@ func (g *GeneratorArgs) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&g.OutputPackagePath, "output-package", "p", g.OutputPackagePath, "Base package path.")
 	fs.StringVarP(&g.OutputFileBaseName, "output-file-base", "O", g.OutputFileBaseName, "Base name (without .go suffix) for output files.")
 	fs.StringVarP(&g.GoHeaderFilePath, "go-header-file", "h", g.GoHeaderFilePath, "File containing boilerplate header text. The string YEAR will be replaced with the current 4-digit year.")
+	fs.BoolVar(&g.GeneratedByComment, "go-generated-by-comment", g.GeneratedByComment, `Add a "Code generated by" comment into each file.`)
```
thockin: do we even need a flag?

sttts: I would drop it. It's something the generator builder decides.

sttts:
> which this pr shouldn't affect.

But it should be replaced with the generic mechanism, right? But if protobuf is happy with `//` we are fine. And if another generator produces code in a language that doesn't like `//`, the comment can be disabled and created manually.

args/args.go
jennybuckley: if we drop the flag how will someone using this by a command line disable it?

jennybuckley: That seems right to me

jennybuckley:
@sttts it should now the comments.

Replacing GENERATOR_NAME with the name of the generator seems to be the most in line with how YEAR is handled.

jennybuckley:
Oh looks like #24 is dropping the ```YEAR``` field entirely

args/args.go
```diff
@@ -103,6 +109,14 @@ func (g *GeneratorArgs) LoadGoBoilerplate() ([]byte, error) {
 		return nil, err
 	}
 	b = bytes.Replace(b, []byte("YEAR"), []byte(strconv.Itoa(time.Now().Year())), -1)
+
+	if g.GeneratedByCommentTemplate != "" {
+		generatorPath := strings.Split(os.Args[0], "/")
+		generatorName := generatorPath[len(generatorPath)-1]
+		generatedByComment = strings.Replace(g.GeneratedByCommentTemplate, "GENERATOR_NAME", generatorName)
+		s := fmt.Sprintf("\n%s\n\n", generatedByComment)
```
sttts: The first newline is not necessary if the file is empty (which it is e.g. in OpenShift)

jennybuckley: I think it could be removed entirely, since the licences usually end with an extra \n

args/args.go
```diff
@@ -103,6 +109,14 @@ func (g *GeneratorArgs) LoadGoBoilerplate() ([]byte, error) {
 		return nil, err
 	}
 	b = bytes.Replace(b, []byte("YEAR"), []byte(strconv.Itoa(time.Now().Year())), -1)
+
+	if g.GeneratedByCommentTemplate != "" {
+		generatorPath := strings.Split(os.Args[0], "/")
+		generatorName := generatorPath[len(generatorPath)-1]
+		generatedByComment = strings.Replace(g.GeneratedByCommentTemplate, "GENERATOR_NAME", generatorName)
```
jennybuckley: should be ```:=``` and needs an extra argument of ```-1```

args/args.go
jennybuckley: Also it would be hard to tell if the the newline should be printed if there is ```// +build !ignore_autogenerated``` above the generated by comment?

jennybuckley:
Looks like I still have to change the example generators to get their outputs to match the test outputs

boilerplate/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright YEAR The Kubernetes Authors.
+Copyright 2017 The Kubernetes Authors.
```
jennybuckley: Pretty sure this is not the right place for this change but since https://github.com/kubernetes/gengo/pull/91 hasn't merged yet, this is necessary to check that the behavior of the example generators isn't affected.

examples/set-gen/main.go
```diff
@@ -35,6 +35,10 @@ import (
 
 func main() {
 	arguments := args.Default()
+
+	// Override defaults.
```
thockin: why?

sttts: @jennybuckley was this only for testing?

jennybuckley: Yes, this was to make the tests pass.

jennybuckley:
@thockin removed the boilerplate change and rebased on #91

examples/set-gen/main.go
thockin: should we fix the tests?

thockin:
#91 merged

examples/set-gen/main.go
jennybuckley: imo, the tests are good how they are because they test if the behavior of the example generators have changed based on the changes to the generic gengo code. I could have run the generators and committed the changes to the generated code, instead of modifying the example generators, but that wouldn't show that the old behavior of the example generators can be implemented by setting the new generator argument, and the output will be exactly how it was before.

args/args.go
```diff
@@ -64,6 +65,11 @@ type GeneratorArgs struct {
 	// Where to get copyright header text.
 	GoHeaderFilePath string
 
+	// If GeneratedByCommentTemplate is set, generate a "Code generated by" comment
+	// above the package declaration of the format defined by this string.
```
sttts: below the boilerplate

args/args.go
```diff
@@ -103,6 +109,14 @@ func (g *GeneratorArgs) LoadGoBoilerplate() ([]byte, error) {
 		return nil, err
 	}
 	b = bytes.Replace(b, []byte("YEAR"), []byte(strconv.Itoa(time.Now().Year())), -1)
+
+	if g.GeneratedByCommentTemplate != "" {
+		generatorPath := strings.Split(os.Args[0], "/")
+		generatorName := generatorPath[len(generatorPath)-1]
```
sttts: https://golang.org/pkg/path/#Base

examples/deepcopy-gen/main.go
```diff
@@ -69,6 +69,7 @@ func main() {
 
 	// Override defaults.
 	arguments.OutputFileBaseName = "deepcopy_generated"
+	arguments.GeneratedByCommentTemplate = "\n// This file was autogenerated by deepcopy-gen. Do not edit it manually!"
```
sttts: Why is this set explicitly. We should use the default.

examples/defaulter-gen/main.go
```diff
@@ -54,6 +54,7 @@ func main() {
 
 	// Override defaults.
 	arguments.OutputFileBaseName = "defaulter_generated"
+	arguments.GeneratedByCommentTemplate = "\n// This file was autogenerated by defaulter-gen. Do not edit it manually!"
```
sttts: we should use the default.

examples/set-gen/main.go
sttts: Yes, please remove the overrides and update the test output in another commit.

jennybuckley:
Hm. now this is failing because of the test case:
```
go run ./examples/set-gen/main.go -i k8s.io/gengo/examples/set-gen/sets/types -o ./examples/set-gen/sets --verify-only
```
Because this generates a binary named "main" instead of "set-gen", it makes the output file differ from the one created by the Makefile. I think this test case is redundant though, because the Makefile already has the same test in it, which passes because it correctly names the binary (it uses go build and then diffs instead of go run and --verify-only). Is there a reason why set-gen is the only one tested with its Makefile and ```go run --verify-only```?

bhcleek:
Can the `Code generated by` comment be made to adhere to format adopted by Go? Doing so would make sure that tooling that respects the officially accepted pattern (e.g. golint) can identify the generated files.

RE:
https://golang.org/s/generatedcode
https://github.com/golang/lint/blob/master/lint.go#L117-L133

jennybuckley:
Rebased on #99

jennybuckley:
@bhcleek I think it does follow that format
```^// Code generated .* DO NOT EDIT\.$```
Where the ```.*``` in the middle is replaced by ```by <gemerator-name>.```

bhcleek:
I agree that it does now. I wrote my comment before your rebase, though, and at the time I wrote it, the comments looked like `This file was autogenerated by deepcopy-gen. Do not edit it manually!`.

Thanks!

sttts:
@thockin @jennybuckley what is blocking this to move forward? We need that for the generated file detection in https://github.com/kubernetes/kubernetes/pull/59172

jennybuckley:
@sttts I think this is able to merge

jennybuckley:
Once this merges, the code generators in k/k will just need to be updated slightly like https://github.com/kubernetes/kubernetes/pull/59674/commits/a34bcd27f3930479cb7621b2f3210e45bc00c3a4

thockin:
@sttts Why are you not in OWNERS for gengo at this point?

Issue #96, opened 2018-01-19
Golint errors in generated files by `deepcopy-gen`


tossmilestone:
When I want to enable golint with `pkg/scheduler` in k8s kubernetes/kubernetes#58234, there are files generated by `deepcopy-gen` having lint errors. e.g.:
```
pkg/scheduler/api/zz_generated.deepcopy.go:265:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
```
The code:
```golang
// DeepCopyObject is an autogenerated deepcopy function, copying the receiver, creating a new runtime.Object.
func (in *Policy) DeepCopyObject() runtime.Object {
	if c := in.DeepCopy(); c != nil {
		return c
	} else {
		return nil
	}
}
```
I think these errors should be fixed in `deepcopy-gen` so these files can pass the golint check.

tossmilestone:
I could fix this issue if it's verified.

thockin:
sounds reasonable.

PR #97, opened 2018-01-19
Fix golint errors in generated files by `deepcopy-gen`.

tossmilestone:
Fix golint errors in generated files by `deepcopy-gen`.

See #96


sttts:
Looks good in general. Please update the output tests to make travis happy.

tossmilestone:
@sttts updated, PTAL.

sttts:
/lgtm

For the record: I wanted to comment "just do `return in.DeepCopy()`". But that's not correct as it return a non-nil interface with a nil pointer inside, instead of a nil-interface. A comment wouldn't be wrong maybe in the generator code.

PR #100, opened 2018-02-20
Add dry run mode

roycaihw:
to allow something like an API linter kubernetes/kubernetes#54887 to use `gengo` without creating any new package.

**NOTE**: for backwards compatibility we have DryRun mode as a separate boolean. Setting it will disable Assemble and Verify modes. 

cc @lavalamp


args/args.go
```diff
@@ -67,6 +67,13 @@ type GeneratorArgs struct {
 	// If true, only verify, don't write anything.
 	VerifyOnly bool
 
+	// If true, only dry run the generator logic, don't write new package or verify
+	// existing package.
+	// NOTE: For backwards compatibility we have DryRun mode as a separate boolean.
+	//       Setting it will disable Assemble and Verify modes, no matter if
+	//       GeneratorArgs.VerifyOnly is true or false.
+	DryRunOnly bool
```
lavalamp: I think this is the wrong place to fix this. You've actually constructed some code which will never generate anything expect (possibly) errors. I can think of two ways to do this:

1. Change your generator to generate a report file rather than a go source file.
2. Change the Generator interface to allow Filename() to be blank, indicating that the generator never writes to a file.

I think 1 is the best option.

lavalamp: Also general thoughts on extending APIs--

Adding a 3rd boolean leaves 3^2 = 8 combinations users can enter. Of those, exactly 3 are valid, 5 are user error. Time to switch to an enum or one-of construct.

"Dry Run" is a confusing concept next to "Verify Only". I think they actually mean the same thing, which is a clue that we're modifying the wrong abstraction. :) Like I said, the thing you want to model is that your package doesn't produce output, not that running the generator doesn't produce output. (Suppose we wanted to run your linter at the same time we run the other generators, which we probably will want to do-- then there's no correct value for this interface change!)

roycaihw: Totally agree on the downside of having another boolean. I had that in mind but implemented it just for backwards compatibility and a quick proof of concept :)

I thought we want a mode for gengo to not generate any file, which the current gengo doesn't support (`Verify` requires packages already being generated). A report file sounds good to me. Closing this pr now

PR #107, opened 2018-06-06
deepcopy: fix []*string and map[string]*string cases

sttts:
Fixes https://github.com/kubernetes/code-generator/issues/45

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -760,8 +760,13 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 				sw.Do("}\n", nil)
 			} else if uet.Kind == types.Pointer {
 				sw.Do("if val==nil { (*out)[key]=nil } else {\n", nil)
-				sw.Do("(*out)[key] = new($.Elem|raw$)\n", uet)
-				sw.Do("val.DeepCopyInto((*out)[key])\n", nil)
+				sw.Do("x := new($.Elem|raw$)\n", uet)
+				sw.Do("(*out)[key] = x\n", uet)
+				if uet.Elem.Kind == types.Builtin {
+					sw.Do("*x = *val\n", nil)
+				} else {
+					sw.Do("val.DeepCopyInto(x)\n", nil)
+				}
 				sw.Do("}\n", nil)
 			} else {
 				sw.Do("(*out)[key] = *val.DeepCopy()\n", uet)
@@ -810,9 +815,14 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
 			sw.Do(fmt.Sprintf("(*out)[i] = (*in)[i].DeepCopy%s()\n", uet.Name.Name), nil)
 			sw.Do("}\n", nil)
 		} else if uet.Kind == types.Pointer {
-			sw.Do("if (*in)[i]==nil { (*out)[i]=nil } else {\n", nil)
-			sw.Do("(*out)[i] = new($.Elem|raw$)\n", uet)
-			sw.Do("(*in)[i].DeepCopyInto((*out)[i])\n", nil)
+			sw.Do("if val := (*in)[i]; val==nil { (*out)[i]=nil } else {\n", nil)
+			sw.Do("x := new($.Elem|raw$)\n", uet)
+			sw.Do("(*out)[i] = x\n", uet)
+			if uet.Elem.Kind == types.Builtin {
+				sw.Do("*x = *val\n", nil)
+			} else {
+				sw.Do("val.DeepCopyInto(x)\n", nil)
+			}
 			sw.Do("}\n", nil)
 		} else if uet.Kind == types.Struct {
 			sw.Do("(*in)[i].DeepCopyInto(&(*out)[i])\n", nil)
--- a/examples/deepcopy-gen/output_tests/maps/doc.go
+++ b/examples/deepcopy-gen/output_tests/maps/doc.go
@@ -22,14 +22,15 @@ package maps
 type Ttest struct {
 	Byte map[string]byte
 	//Int8    map[string]int8 //TODO: int8 becomes byte in SnippetWriter
-	Int16   map[string]int16
-	Int32   map[string]int32
-	Int64   map[string]int64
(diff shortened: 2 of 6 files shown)


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -761,7 +761,12 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 			} else if uet.Kind == types.Pointer {
 				sw.Do("if val==nil { (*out)[key]=nil } else {\n", nil)
 				sw.Do("(*out)[key] = new($.Elem|raw$)\n", uet)
-				sw.Do("val.DeepCopyInto((*out)[key])\n", nil)
+				if uet.Elem.Kind == types.Builtin {
+					sw.Do("x := *val\n", nil)
```
thockin: Won't this kind of trash Go's concept of heap and stack?  I know Go is smart enough to move things but ISTR that the general guidance for perf was to avoid t if you can.  In this case, maybe we should use `new` ?

examples/deepcopy-gen/output_tests/slices/zz_generated.go
```diff
@@ -78,6 +78,19 @@ func (in *Ttest) DeepCopyInto(out *Ttest) {
...
 		copy(*out, *in)
 	}
+	if in.StringPtr != nil {
+		in, out := &in.StringPtr, &out.StringPtr
+		*out = make([]*string, len(*in))
+		for i := range *in {
+			if (*in)[i] == nil {
+				(*out)[i] = nil
+			} else {
+				(*out)[i] = new(string)
```
lavalamp: This line is a needless allocation.

examples/deepcopy-gen/generators/deepcopy.go
thockin: Also, we used `new` on L763, can we just assign into that value here

examples/deepcopy-gen/output_tests/slices/zz_generated.go
```diff
@@ -78,6 +78,19 @@ func (in *Ttest) DeepCopyInto(out *Ttest) {
...
+				x := *(*in)[i]
+				(*out)[i] = &x
```
lavalamp: If you want to keep the new (per Tim) then these two lines should just be this one: ```*(*out)[i] = *(*in)[i]```

sttts: fixed

examples/deepcopy-gen/output_tests/slices/zz_generated.go
sttts: fixed

examples/deepcopy-gen/generators/deepcopy.go
sttts: fixed

examples/deepcopy-gen/output_tests/maps/doc.go
```diff
@@ -22,14 +22,15 @@ package maps
...
+	Int32     map[string]int32
+	Int64     map[string]int64
+	Uint8     map[string]uint8
+	Uint16    map[string]uint16
+	Uint32    map[string]uint32
+	Uint64    map[string]uint64
+	Float32   map[string]float32
+	Float64   map[string]float64
+	String    map[string]string
+	StringPtr map[string]*string
```
lavalamp: Hm, does `map[string]**string` work too?

examples/deepcopy-gen/output_tests/maps/zz_generated.go
```diff
@@ -100,6 +100,18 @@ func (in *Ttest) DeepCopyInto(out *Ttest) {
...
 	}
+	if in.StringPtr != nil {
+		in, out := &in.StringPtr, &out.StringPtr
+		*out = make(map[string]*string, len(*in))
+		for key, val := range *in {
+			if val == nil {
+				(*out)[key] = nil
+			} else {
+				(*out)[key] = new(string)
+				*(*out)[key] = *val
```
lavalamp: Unless go manages to optimize it, this is a lot of map deference operations?

```go
x := new(string)
*x = *val
(*out)[key] = x
```

might be fewer?

Quite probably, not worth changing--up to you.

examples/deepcopy-gen/output_tests/maps/doc.go
thockin: might as well add tests cases

examples/deepcopy-gen/output_tests/maps/zz_generated.go
thockin: might be easier to read

sttts: will optimize that.

examples/deepcopy-gen/output_tests/maps/doc.go
sttts: I doubt ** works. But how often have you seen ** in Go? With aliases this might get messy, if we come up with something more generic.

thockin:
I agree we don't NEED it but it should work.  If it doesn't, it just means
our generator has a bug.  It may be a P4 bug...

examples/deepcopy-gen/output_tests/maps/zz_generated.go
sttts: fixed

thockin:
Travis is grumpy with a type error.

sttts:
> Travis is grumpy with a type error.

Never rename something quickly before push.

examples/deepcopy-gen/output_tests/maps/doc.go
lavalamp: If it doesn't work the deep copier needs to at least fail rather than write wrong code.

lavalamp: (in a follow-up is fine)

lavalamp:
/lgtm
/approve

Can you file an issue about `**` if you don't want to fix it now?

examples/deepcopy-gen/output_tests/maps/doc.go
sttts: Created an issue https://github.com/kubernetes/gengo/issues/109

We create uncompilable code at least, though proper errors would be certainly much better (true for many of our generators).

Issue #109, opened 2018-06-07
Deepcopy broken for deeper nested pointers of builtins as slice+map value type


sttts:
For `[]**string`, `[]***int`, `map[string]**bool` etc. we generate deepcopy code which calls `DeepCopyInto`, which is not defined for those types.

xref https://github.com/kubernetes/gengo/pull/107#discussion_r193825992

PR #110, opened 2018-06-08
deepcopy: make slice+map generation uniform and nestable

sttts:
This PR uses `g.generateFor` inside slice and map deepcopy generation instead of special casing of some non-nested types. `g.generateFor` is supposed to be nestable such that we can generate code for arbitrarily nested types like  `[]*map[string]***[]*string`.

Moreover, this PR
- adds reflection deepcopy unit tests of the output
- adds a benchmark
- makes the generator fail early and hard when there is an unknown type case, instead of best effort code generation with `out = in.DeepCopy()`.

Fixes https://github.com/kubernetes/gengo/issues/109.


sttts:
Due to the slightly more complex generated map code, this deserved a benchmark.

lavalamp:
The expanded test case looks awesome. This LGTM if you squash

sttts:
@lavalamp will add a benchmark and give @thockin a chance to look at the updated output per commit. Will squash then at the end.

sttts:
And I switched to hard errors instead of best effort `out = in.DeepCopy()` generation. Want to test-vendor into kube to see whether we missed cases.

examples/deepcopy-gen/output_tests/maps/zz_generated.go
```diff
@@ -104,13 +104,152 @@ func (in *Ttest) DeepCopyInto(out *Ttest) {
...
 		for key, val := range *in {
+			var outVal *string
 			if val == nil {
 				(*out)[key] = nil
 			} else {
-				x := new(string)
-				(*out)[key] = x
-				*x = *val
+				in, out := &val, &outVal
+				if *in == nil {
```
thockin: This is redundant with L107/108 above

sttts:
Added deepcopy tests with fuzzing, comparing against reflection based deepcopy.

And the benchmarks, before the structural changes:

```
300000	      4655 ns/op
3000000	       588 ns/op
20000000	       107 ns/op
```

and after

```
300000	      4509 ns/op
3000000	       584 ns/op
20000000	       106 ns/op
```

examples/deepcopy-gen/output_tests/maps/zz_generated.go
sttts: fixed. Pointers had a redundant nil check.

sttts:
@lavalamp @thockin squashed. ptal.

PR #112, opened 2018-06-24
Passing a path of "" is undefined for different operating systems. 

taliesins:
Passing a path of "" is undefined for different operating systems. To make the behavior consistent with Linux pass in the current working directory.

This does not work on Windows due to: https://github.com/golang/go/issues/24441

### DIFF
--- a/generator/execute.go
+++ b/generator/execute.go
@@ -159,7 +159,11 @@ func assembleGolangFile(w io.Writer, f *File) {
 }
 
 func importsWrapper(src []byte) ([]byte, error) {
-	return imports.Process("", src, nil)
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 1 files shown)


generator/execute.go
```diff
@@ -159,7 +159,11 @@ func assembleGolangFile(w io.Writer, f *File) {
 }
 
 func importsWrapper(src []byte) ([]byte, error) {
-	return imports.Process("", src, nil)
+	dir, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	return imports.Process(dir, src, nil)
```
thockin: Process expects a file name to do path-relative resolution (e.g. nested vendor/).  How hard would it be to plumb that down?

@lavalamp @sttts

sttts: Compare https://github.com/kubernetes/gengo/pull/115. Answer: simple.

PR #113, opened 2018-06-29
Follow pkg alias convention of kube of not having _ between dirs

sttts:
We are starting excluding certain generated files from linting and verify scripts because we use underscores in import aliases. Instead, we should fix Gengo here to comply with the kube conventions.

Compare https://github.com/kubernetes/kubernetes/pull/64664/files#diff-4175e534855bc6fa66d08dee56797817R29


MHBauer:
finally
/lgtm

PR #114, opened 2018-06-30
Sanitize pkg names containing _ with underscore-less alias

sttts:
Package names like pkg/apis/custom_metrics were still aliased with custom_metrics. After this PR the alias is custommetrics.

### DIFF
--- a/generator/import_tracker.go
+++ b/generator/import_tracker.go
@@ -49,6 +49,7 @@ func golangTrackerLocalName(tracker namer.ImportTracker, t types.Name) string {
 	for n := len(dirs) - 1; n >= 0; n-- {
 		// follow kube convention of not having anything between directory names
 		name := strings.Join(dirs[n:], "")
+		name = strings.Replace(name, "_", "", -1)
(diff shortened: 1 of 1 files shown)


generator/import_tracker.go
```diff
@@ -49,6 +49,7 @@ func golangTrackerLocalName(tracker namer.ImportTracker, t types.Name) string {
 	for n := len(dirs) - 1; n >= 0; n-- {
 		// follow kube convention of not having anything between directory names
 		name := strings.Join(dirs[n:], "")
+		name = strings.Replace(name, "_", "", -1)
```
thockin: I am pedantically worried about collisions, but in practice it seems unlikely.  Any ideas on how to make this less ambiguous?

sttts: Valid concern. But as we don't have many (quite the opposite) underscores in package names, the chances are low. And the chance of conflicts IMO is much much higher for single word package names and symbol names which do not go through a namer. E.g. let's hope nobody calls a package `in` or `out` with imported types from there.

PR #117, opened 2018-07-14
fixes, returns an empty string when the name length is less than 2

wzshiming:


wzshiming:
cla

namer/plural_namer.go
```diff
@@ -59,7 +59,7 @@ func (r *pluralNamer) Name(t *types.Type) string {
 		return r.finalize(plural)
 	}
 	if len(singular) < 2 {
-		return r.finalize(plural)
+		return r.finalize(singular)
```
sttts: Looks right. Can we have a unit test for this?

namer/plural_namer_test.go
```diff
@@ -35,6 +35,11 @@ func TestPluralNamer(t *testing.T) {
 		expectedPrivate string
 		expectedPublic  string
 	}{
+		{
+			"I",
+			"i",
```
sttts: Does our apimachinery/apiserver code handle these well? Would have expected `I`, `is`, `Is`.

wzshiming: The fact that a name is a single letter has yet to appear.
I found it and thought it was a problem.

sttts: I just want to avoid that we try to fix it in a way that it will break in apimachinery/apiserver. We better double check that the solution works throughout the stack.

wzshiming: I check that the works throughout the stack, this is only used in the code-generator, does not affect any code until it is regenerated. If there is a problem after the regenerate, the compilation will not pass.

sttts: No, I mean that it is used e.g. in clientsets. If you define a type `type A struct {...}`, e.g. a CRD or an aggregated API server. Will both work?

wzshiming: In working order.

sttts: To test this, vendor your change into kube and then add some test cases with one-letter types and package names into https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/code-generator/_examples/crd/apis.

wzshiming: I submitted test cases. 
https://github.com/kubernetes/kubernetes/pull/66436

sttts:
@wzshiming please revendor this into your other PR.

PR #121, opened 2018-08-08
deepcopy: unify and fix first and second closed comment logic

sttts:
Fixes `// +k8s:deepcopy=true` being ignore above a whitespace line.

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -40,33 +40,37 @@ type CustomArgs struct {
 
 // This is the comment tag that carries parameters for deep-copy generation.
 const (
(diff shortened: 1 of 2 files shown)


thockin:
All of this logic around closest comments and second closest comments is somewhat inane, don't you think?  Is it worthwhile to think of a better mental model for this?  As a human, it's pretty clear what we want, but less easy to express in code..

sttts:
> All of this logic around closest comments and second closest comments is somewhat inane, don't you think? Is it worthwhile to think of a better mental model for this? As a human, it's pretty clear what we want, but less easy to express in code..

Actually, I wouldn't mind to just merge all comments between code into one block and remove all this distinction in gengo. Do we have real cases where this is not what the user expects?

sttts:
Rebased due to klog changes.

PR #122, opened 2018-08-08
go-to-protobuf: only make named types with package global

sttts:
Otherwise, the first occurance of e.g. "*string" will be made global, introducing import dependencies to totally unrelated package in the produced .proto files.


sttts:
Moved to https://github.com/kubernetes/kubernetes/pull/67158

PR #125, opened 2018-09-29
report all import errors one time

bruceauyeung:
without this PR, `hack/verify-import-boss.sh` reports errors one by one (for example report only one forbidden error even if the package have more than one imports match forbidden prefixes )

when i add an import restriction, i have to run `hack/verify-import-boss.sh`,and then add an allowed prefix, and then run `hack/verify-import-boss.sh`,and then add an allowed prefix, over and over again.
it's really a waste of time.

### DIFF
--- a/examples/import-boss/generators/import_restrict.go
+++ b/examples/import-boss/generators/import_restrict.go
@@ -19,6 +19,7 @@ package generators
 
 import (
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


bruceauyeung:
@lavalamp @wojtek-t  
hi, would you plz take a look ? thanks!

PR #129, opened 2018-11-04
Switch from glog to klog

dims:
Change-Id: I0a8ec9a038129690d594f8dbc6a1acd611a31659


dims:
related to https://github.com/kubernetes/kubernetes/issues/70264

Issue #133, opened 2019-01-17
Deep copy gen generates invalid code if struct has an `_` field


cholman-zd:
Running deepcopy-gen against structs that have `_` fields in them generates invalid go code.

Source struct:

```
type CreateBucketConfiguration struct {
	_ struct{} `type:"structure"`

	LocationConstraint BucketLocationConstraint `type:"string" enum:"true"`
}
type BucketLocationConstraint string
```

`doc.go` directives to deepcopy:

```
// +k8s:deepcopy-gen=package,register
```

deepcopy directives above structs that the above struct is a child of  (not sure if this makes a difference or not - this is autogenerated code from the operator sdk)

```
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```

Output by deepcopy-gen:
```
// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
func (in *CreateBucketConfiguration) DeepCopyInto(out *CreateBucketConfiguration) {
	*out = *in
	out._ = in._
	return
}
```

Setting or accessing `_` fields is invalid. 

I applied the following patch locally and it outputs correct code (ie does not attempt to copy the `_` field).  No idea if this is the _correct_ way to solve the problem though.

```
diff --git a/vendor/k8s.io/gengo/examples/deepcopy-gen/generators/deepcopy.go b/vendor/k8s.io/gengo/examples/deepcopy-gen/generators/deepcopy.go
index 4548108..1d90578 100644
--- a/vendor/k8s.io/gengo/examples/deepcopy-gen/generators/deepcopy.go
+++ b/vendor/k8s.io/gengo/examples/deepcopy-gen/generators/deepcopy.go
@@ -823,6 +823,9 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {

 	// Now fix-up fields as needed.
 	for _, m := range ut.Members {
+		if m.Name == "_" {
+			continue
+		}
 		ft := m.Type
 		uft := underlyingType(ft)
```

After applying this patch, output is as expected:

```
// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
func (in *CreateBucketConfiguration) DeepCopyInto(out *CreateBucketConfiguration) {
	*out = *in
	return
}
```

cholman-zd:
[...]

PR #134, opened 2019-02-01
Fix Code gen bug that generates invalid code if struct has _.

kunday:
Fixes bug - https://github.com/kubernetes/gengo/issues/133 

The Issue has full detail from @cholman-zd and this one just lifts the ability to skip `_` fields so the codegen does not hiccup.

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -823,6 +823,9 @@ func (g *genDeepCopy) doStruct(t *types.Type, sw *generator.SnippetWriter) {
 
 	// Now fix-up fields as needed.
 	for _, m := range ut.Members {
+		if m.Name == "_" {
+			continue
+		}
 		ft := m.Type
 		uft := underlyingType(ft)
 


cholman-zd:
Thanks @kunday , lgtm!

sttts:
We need a test for that. Please add it to the example files.

kunday:
@stts There are no existing tests in the examples/deepcopy-gen that tests this execution. 

Do you have any suggestions on how to go about this?

lavalamp:
Add an example here: https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/output_tests/structs/doc.go

epssy:
Every time I forget to include this patch my build fails.  It's frustrating.  

@lavalamp could you please give some more guidance on what sort of test you're looking for and I'll write something.  If you could be somewhat detailed as I found the link to `doc.go` to be rather confusing when I had a look last.

grim-fendango:
Bump

lavalamp:
The tests in https://github.com/kubernetes/gengo/tree/master/examples/deepcopy-gen/output_tests are just go code and corresponding output. You need to add an example (probably in the struct package I linked to) and the corresponding change in the output.

Issue #138, opened 2019-03-06
deepcopy-gen generates invalid code for interface{}


j-vizcaino:
Declaring a `map[string]interface{}` member in a struct handled by deepcopy-gen generates invalid code.
It seems there could be two options available:

* for `map[string]interface{}`, generate code that leverages `runtime.DeepCopyJSON`
* for any `interface{}`, error out instead of generating code that does not compile.

Related issues: #128, https://github.com/kubernetes-sigs/kubebuilder/issues/528 and https://github.com/kubernetes-sigs/controller-tools/pull/126

### Example

* `foo.go`

```golang
// +k8s:deepcopy-gen=true
type Foo struct {
  Data map[string]interface{}
}
```

* `zz_generated.deepcopy.go` (`DeepCopy` function omitted for clarity as it basically calls `DeepCopyInto`)

```golang
// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
func (in *Foo) DeepCopyInto(out *Foo) {
	*out = *in
	if in.Data != nil {
		in, out := &in.Data, &out.Data
		*out = make(map[string]interface{}, len(*in))
		for key, val := range *in {
			if val == nil {
				(*out)[key] = nil
			} else {
				// INVALID FUNCTION NAME
				(*out)[key] = val.DeepCopyinterface{}()
			}
		}
	}
	return
}
```

sttts:
Please give an example of the code which is wrong now.

j-vizcaino:
@sttts Updated the issue description with an example.

sttts:
It's kind of intentional, though I agree we should show an error in deepcopy-gen instead of `DeepCopyinterface{}` (PR welcome)

In the case of named interfaces, the code is correct:
```
func (in *Foo) DeepCopyInto(out *Foo) {
	*out = *in
	if in.Data != nil {
		in, out := &in.Data, &out.Data
		*out = make(map[string]interface{}, len(*in))
		for key, val := range *in {
			if val == nil {
				(*out)[key] = nil
			} else {
				// INVALID FUNCTION NAME
				(*out)[key] = val.DeepCopyNamedInterface()
			}
		}
	}
	return
}
```

Use `// +k8s:deepcopy-gen:interfaces=<package-path>.NamedIntrerface` to implement `DeepCopyNamedInterface`.

j-vizcaino:
Yes, for named type this works properly (otherwise nothing would work, I presume 🤔).

[...]

PR #140, opened 2019-03-18
deepcopy: error out on interface{}

j-vizcaino:
Instead of generating code that does not compile, error out immediately
when generateFor gets called with `interface{}`

This relates to #138 

This PR lacks testing for this: I'll need some guidance around this please.

I currently see two ways to do this:

* factorize the code that handles the interface type and unit-test this function
* add another type of end-to-end test, this time making sure that `deepcopy-gen` fails. This one seems a bit brittle as the cause of failure would need to be checked for the test to be effective.

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -745,6 +745,10 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 	case uet.IsAssignable():
 		sw.Do("(*out)[key] = val\n", nil)
 	case uet.Kind == types.Interface:
+		// Note: do not generate code that won't compile as `DeepCopyinterface{}()` is not a valid function
+		if uet.Name.Name == "interface{}" {
+			klog.Fatalf("DeepCopy of %q is unsupported. Instead, use named interfaces with DeepCopy<named-interface> as one of the methods.", uet.Name.Name)
+		}
 		sw.Do("if val == nil {(*out)[key]=nil} else {\n", nil)
 		// Note: if t.Elem has been an alias "J" of an interface "I" in Go, we will see it
 		// as kind Interface of name "J" here, i.e. generate val.DeepCopyJ(). The golang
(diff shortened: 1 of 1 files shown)


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -745,6 +745,10 @@ func (g *genDeepCopy) doMap(t *types.Type, sw *generator.SnippetWriter) {
 	case uet.IsAssignable():
 		sw.Do("(*out)[key] = val\n", nil)
 	case uet.Kind == types.Interface:
+		// Note: do not generate code that won't compile as `DeepCopyinterface{}()` is not a valid function
+		if uet.Name.Name == "interface{}" {
+			klog.Fatalf("Hit an unsupported type %q", uet.Name.Name)
```
lavalamp: I think I'd expect a runtime check as to whether it implements the deep copy interface rather than a hard fail?

@sttts

j-vizcaino: @lavalamp Can you describe a bit more precisely what you have in mind please? I'm not sure I understand your comment.

sttts: @lavalamp I think the change is fine. For every other, named interface you get proper output from the compiler which tells you that you have to implement the deep copy func. For `interface{}` though we get syntactical garbage. We could improve the former case too, but that's a different issue worth a follow-up (with limited value).

examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -793,6 +797,10 @@ func (g *genDeepCopy) doSlice(t *types.Type, sw *generator.SnippetWriter) {
 			g.generateFor(ut.Elem, sw)
 			sw.Do("}\n", nil)
 		} else if uet.Kind == types.Interface {
+			// Note: do not generate code that won't compile as `DeepCopyinterface{}()` is not a valid function
+			if uet.Name.Name == "interface{}" {
+				klog.Fatalf("Hit an unsupported type %q", uet.Name.Name)
```
sttts: can we improve the output? Not sure we have any helpful context here (we really should have something like a path through the types). Maybe `DeepCopy of interface{} is unsupported. Instead, use named interfaces with DeepCopy<named-interface> as one of the methods.`.

j-vizcaino: Good call! Updated the PR with the proposed wording. You're right, it provides a better guidance for users hitting this issue.

examples/deepcopy-gen/generators/deepcopy.go
lavalamp: SGTM

Issue #141, opened 2019-03-26
CI is silently failing


andreykaipov:
Running the current first step of the build shows the issue:
```
$ find . -name vendor -prune -o -name Makefile -execdir make test \;
unknown flag: --logtostderr
Usage of /tmp/defaulter-gen:
...
make: *** [test] Error 2
...
```

This started happening 5 months when the repo switched over from glog to klog (https://github.com/kubernetes/gengo/pull/129). The build log for it is https://travis-ci.org/kubernetes/gengo/jobs/450604661.

Since `find` wraps the error and returns zero, the CI has been silently failing since.

PR #142, opened 2019-03-26
Fix silently failing CI

andreykaipov:
Fixes https://github.com/kubernetes/gengo/issues/141

The examples needed to initialize klog's flags. Also we make sure the build returns non-zero on errors so this can be caught sooner in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jobs:
   include:
     - stage: Run tests
       script:
-        - find . -name vendor -prune -o -name Makefile -execdir make test \;
+        - find . -name vendor -prune -o -name Makefile -print | xargs -I% sh -c 'make -C $(dirname %) test'
         - go test -v ./...
     - stage: Verify examples
       script:
--- a/examples/deepcopy-gen/main.go
+++ b/examples/deepcopy-gen/main.go
@@ -65,6 +65,7 @@ import (
 )
 
 func main() {
+	klog.InitFlags(nil)
 	arguments := args.Default()
 
 	// Override defaults.
--- a/examples/defaulter-gen/main.go
+++ b/examples/defaulter-gen/main.go
@@ -50,6 +50,7 @@ import (
 )
 
 func main() {
+	klog.InitFlags(nil)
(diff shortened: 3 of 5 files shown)


.travis.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   include:
     - stage: Run tests
       script:
-        - find . -name vendor -prune -o -name Makefile -execdir make test \;
+        - find examples -name Makefile -print | xargs -I% sh -c 'make -C $(dirname %)'
```
sttts: this does not run the test target anymore

.travis.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   include:
     - stage: Run tests
       script:
-        - find . -name vendor -prune -o -name Makefile -execdir make test \;
+        - find examples -name Makefile -print | xargs -I% sh -c 'make -C $(dirname %) test'
```
lavalamp: It seems more future-proof to search '.' for Makefiles.

Side note: when did this repo grow Makefiles??

andreykaipov: I changed it to search `examples` rather than `.` since `examples` is the only dir with any Makefiles currently. It feels odd searching for any Makefile only to have to prune out the directories we don't want (in this case the `vendor` dir).

I imagine if somebody wants to add an extra Makefile in the future to some other directory,  they'll modify this script to include the new dir. This way testing on the CI is opt-in, rather than opt-out. Does explaining the intention change your mind, or is that opposite of what you meant by future-proofing? 

And I can't answer your second question! 😅 I only caught this because I wanted to submit a feature PR but wasn't comfortable with the failing tests so I wanted to fix them first.

sttts: I tend to agree with @lavalamp, i.e. pruning is better as it avoid that somebody adds a non-example Makefile which is silently ignored.

andreykaipov: Okiedoke - I've updated it!

lavalamp:
Looks good, please squash and we'll get this merged.

andreykaipov:
Squashed! 😄

Issue #143, opened 2019-04-04
Update for go modules


lavalamp:
Let's begin publishing the module file.

Let's version this thing and start respecting the semver guarantees.

lavalamp:
I'd like to tag v1.0.0 after merging #144. Any objections? I think this commits us to not breaking go interfaces. @sttts @thockin

liggitt:
I think we need a tool to tell us if we're breaking go interfaces or we'll fail at manual inspection/enforcement

I assume this applies not only to the code in this repo but also to the code generated by what is in this repo?

Is the plan to use semantic import tagging for v2 and up?

lavalamp:
good point, I thought I heard something about the go team providing such a tool?

I guess we should also not break the code the samples generate. Maybe we should put the samples in different modules. If we check in the generated code, we could use the above tool to verify this.

> Is the plan to use semantic import tagging for v2 and up?

I am not sure I understand this question, I thought for v2 we actually have to have a v2 in the path, but for v1 we shouldn't. I wasn't planning on thinking about this until someone wants to break the interface :)

liggitt:
> I thought I heard something about the go team providing such a tool?

https://github.com/golang/go/tree/master/api was all I found, but it seems hard-coded to golang src, so would require forking to use against something else

> I am not sure I understand this question, I thought for v2 we actually have to have a v2 in the path, but for v1 we shouldn't.

correct

> I wasn't planning on thinking about this until someone wants to break the interface :)

fair

liggitt:
> good point, I thought I heard something about the go team providing such a tool?

I also saw https://github.com/smola/gocompat mentioned (https://github.com/kubernetes/client-go/issues/234#issuecomment-445785653), but ran into issues trying to run it on our repos

liggitt:
I think https://github.com/golang/tools/tree/master/internal/apidiff is the tool in the works from the go team

maelvls:
a

PR #144, opened 2019-04-04
initialize go module file

lavalamp:


lavalamp:
Half of #143.

lavalamp:
Ran `go mod tidy` & updated.

liggitt:
should this bump Travis.yml to go1.12 and run with GO111MODULE=on to make sure CI is actually testing with the versions listed here?

andreykaipov:
Oh I was recently trying to get gengo to use Go modules too! 😄

To get the tests to pass (specifically `TestImportBuildPackage` from `parser/local_parse_test.go`), we have to add an empty `go.mod` to `vendor/fake/dep`:
```
module _

go 1.12
```

And reference it from the root `go.mod`:
```
module k8s.io/gengo

go 1.12

require (
        fake/dep v0.0.0
        ...
)

replace fake/dep v0.0.0 => ./vendor/fake/dep
```
In addition, because of an assumption the library took for importing packages (see [parser/parse.go#551](https://github.com/kubernetes/gengo/blob/master/parser/parse.go#L551)), we have to use `$(go list k8s.io/gengo/...)` instead of `k8s.io/gengo/...` directly for the second stage of the CI when verifying the examples (see [.travis.yml#L16](https://github.com/kubernetes/gengo/blob/master/.travis.yml#L16)).

The long-term solution would be to migrate from `go/build` to `golang.org/x/tools/go/packages` as recommended by the Go 1.11 release notes https://golang.org/doc/go1.11#gopackages. The way I understand it is by invoking `go list` ourselves, we're essentially doing what `go/packages` would do for us.

lavalamp:
@andreykaipov Sure, that'd be great!

estroz:
@andreykaipov are you working on this? If not I can take a stab at it.

andreykaipov:
Hi @estroz - I'm not working on it. At the time of my comment I tried to do it but couldn't really wrap my head around it. :sweat_smile:

The only PR I have is the one above into this branch to get the CI working with Go modules, but that might have to change if `go/build` is replaced.

PR #150, opened 2019-08-13
 Sets: Make Insert and Delete return the Set

alvaroaleman:
Sets: Make Insert and Delete return the Set
    
Sets in its current form is already very useful, however sometimes it is
awkward to use. Consider the following example:
    
```
// Removes a finalizer if it exists
finalizers := sets.NewString(obj.GetFinalizers()...)
sets.Delete(toRemove)
obj.SetFinalizers(set.List())
```
    
With this patch, the above example could be solved in one line:
    
```
obj.SetFinalizers(sets.NewString(obj.GetFinalizers()...).Delete(toDelete).List())
```

Result of discussion in https://github.com/kubernetes/kubernetes/pull/81234

/assign @lavalamp

### DIFF
--- a/examples/set-gen/generators/sets.go
+++ b/examples/set-gen/generators/sets.go
@@ -205,17 +205,19 @@ func $.type|public$KeySet(theMap interface{}) $.type|public$ {
 }
 
 // Insert adds items to the set.
-func (s $.type|public$) Insert(items ...$.type|raw$) {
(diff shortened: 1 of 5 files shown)


lavalamp:
/lgtm
/approve

Technically this is a breaking change if folks are capturing the methods in variables or something, but I really don't expect usage like that to be common. So, apologies to anyone bitten by this; I think it's worth it overall.

PR #151, opened 2019-08-22
Optionally allow processing of *_test.go files

dims:
Currently we skip any test related go files, for a specific scenario
like verify-import-boss in upstream kubernetes, we would like to verify
the imports of test code in addition as well. So it would be great to
add a flag to specify that we need to process them as well.


dims:
Please see https://github.com/kubernetes/kubernetes/pull/81752 for PoC

parser/parse.go
```diff
@@ -304,18 +307,25 @@ func (b *Builder) addDir(dir string, userRequested bool) error {
...
-		data, err := ioutil.ReadFile(absPath)
-		if err != nil {
-			return fmt.Errorf("while loading %q: %v", absPath, err)
-		}
-		err = b.addFile(pkgPath, absPath, data, userRequested)
-		if err != nil {
-			return fmt.Errorf("while parsing %q: %v", absPath, err)
+	directories := [][]string{buildPkg.GoFiles}
+	if b.IncludeTestFiles {
+		directories = append(directories, buildPkg.TestGoFiles)
```
sttts: why so complicated? just use a `files := []string{}` and append `GoFiles...` and `TestGoFiles...`.

dims:
Thanks @sttts fixed.

PR #116, opened 2018-07-10
import-boss: add inverse import rules

sttts:
Allow to restrict which packages Y may import a given package X via `InverseRules` in `X/.import-restrictions`:

```
//   "InverseRules": [{
//       "SelectorRegexp": "k8s[.]io",
//       "AllowedPrefixes": [
//         "k8s.io/same-repo",
//         "k8s.io/kubernetes/pkg/legacy"
//       ],
//       "ForbiddenPrefixes": [
//         "k8s.io/kubernetes/pkg/legacy/subpkg"
//       ]
//     },
//     {
//       "SelectorRegexp": "k8s[.]io",
//       "Transitive": true,
//       "AllowedPrefixes": [
//         "k8s.io/
//       ],
//       "ForbiddenPrefixes": [
//         "k8s.io/kubernetes/cmd/kubelet",
//         "k8s.io/kubernetes/cmd/kubectl"
//       ],
//   ]
// }
```

Example use-cases:
- k8s.io/apiserver/pkg/features feature gates should not leak into kubectl and kubelet (e.g. for https://github.com/kubernetes/kubernetes/pull/65558).
- k8s.io/kubernetes/pkg/master should only be imported transitively by cmd/kube-apiserver and cmd/hyperkube.
- k8s.io/apimachinery/pkg/private should only be imported directly by k8s.io/apimachinery, with the legacy exceptions a, b, c.

PR #154, opened 2019-09-25
import-boss: add inverse import rules

eloyekunle:
Continues from #116 by @sttts.

Closes #116

### DIFF
--- a/examples/import-boss/generators/import_restrict.go
+++ b/examples/import-boss/generators/import_restrict.go
@@ -38,6 +38,7 @@ import (
 )
 
 const (
+	goModFile          = "go.mod"
 	importBossFileType = "import-boss"
 )
 
@@ -58,7 +59,7 @@ func DefaultNameSystem() string {
 func Packages(c *generator.Context, arguments *args.GeneratorArgs) generator.Packages {
 	pkgs := generator.Packages{}
 	c.FileTypes = map[string]generator.FileType{
-		importBossFileType: importRuleFile{},
+		importBossFileType: importRuleFile{c},
 	}
 
 	for _, p := range c.Universe {
@@ -70,6 +71,7 @@ func Packages(c *generator.Context, arguments *args.GeneratorArgs) generator.Pac
 		pkgs = append(pkgs, &generator.DefaultPackage{
 			PackageName: p.Name,
 			PackagePath: p.Path,
+			Source:      p.SourcePath,
 			// GeneratorFunc returns a list of generators. Each generator makes a
 			// single file.
 			GeneratorFunc: func(c *generator.Context) (generators []generator.Generator) {
@@ -96,10 +98,19 @@ type Rule struct {
 	ForbiddenPrefixes []string
 }
 
+type InverseRule struct {
+	Rule
+	// True if the rule is to be applied to transitive imports.
+	Transitive bool
+}
+
 type fileFormat struct {
 	CurrentImports []string
 
-	Rules []Rule
+	Rules        []Rule
+	InverseRules []InverseRule
+
+	path string
 }
 
 func readFile(path string) (*fileFormat, error) {
@@ -113,6 +124,7 @@ func readFile(path string) (*fileFormat, error) {
 	if err != nil {
 		return nil, fmt.Errorf("couldn't unmarshal %v: %v", path, err)
 	}
+	current.path = path
 	return &current, nil
 }
 
@@ -131,10 +143,12 @@ func writeFile(path string, ff *fileFormat) error {
 }
 
 // This does the actual checking, since it knows the literal destination file.
-type importRuleFile struct{}
+type importRuleFile struct {
+	context *generator.Context
+}
 
-func (importRuleFile) AssembleFile(f *generator.File, path string) error {
-	return nil
+func (irf importRuleFile) AssembleFile(f *generator.File, path string) error {
+	return irf.VerifyFile(f, path)
 }
 
 // TODO: make a flag to enable this, or expose this information in some other way.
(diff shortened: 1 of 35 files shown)


examples/import-boss/main.go
```diff
@@ -16,13 +16,15 @@ limitations under the License.
 
 // import-boss enforces import restrictions in a given repository.
 //
-// When a directory is verified, import-boss looks for a file called
-// ".import-restrictions". If this file is not found, parent directories will be
-// recursively searched.
+// When a package is verified, import-boss looks for files called
+// ".import-restrictions" in all directories between the package and
+// the $GOPATH/src.
```
sttts: what does this mean if we become `$GOPATH`-less using go mod?

examples/import-boss/generators/import_restrict.go
```diff
@@ -266,6 +279,65 @@ func (irf importRuleFile) verifyRules(restrictionFiles []*fileFormat, f *generat
 	return nil
 }
 
+func (irf importRuleFile) verifyInverseRules(restrictionFiles []*fileFormat, f *generator.File, path string) error {
+	selectors := make([][]*regexp.Regexp, len(restrictionFiles))
```
sttts: can you add a short comment above each block with what it does? like a one- or twoliner.

examples/import-boss/generators/import_restrict.go
```diff
@@ -266,6 +279,65 @@ func (irf importRuleFile) verifyRules(restrictionFiles []*fileFormat, f *generat
 	return nil
 }
 
+func (irf importRuleFile) verifyInverseRules(restrictionFiles []*fileFormat, f *generator.File, path string) error {
```
sttts: godoc for the func would be nice

examples/import-boss/main.go
```diff
@@ -55,11 +61,37 @@ limitations under the License.
...
+//       ],
+//       "ForbiddenPrefixes": [
+//         "k8s.io/kubernetes/cmd/kubelet",
+//         "k8s.io/kubernetes/cmd/kubectl"
+//       ],
 //   ]
 // }
 //
-// Note the second block explicitly matches the unsafe package, and forbids it
+// Note the second rule explicitly matches the unsafe package, and forbids it
```
sttts: second (non-inverse) rule

generator/transitive_closure.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
sttts: let's update the dates to 2019

generator/transitive_closure.go
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+package generator
+
+type edge struct {
+	from string
+	to   string
+}
+
+func transitiveClosure(in map[string][]string) map[string][]string {
```
sttts: this deserves a unit test

sttts:
/cc @dims 

This will allow us to disallow imports of packages in place we don't want, within the monorepo. Like `internal` packages from Golang, but much more flexible. I.e. it allows us to put up import barriers long before a package is actually moved into a staging repo. For example we could restrict that internal API types are only used inside of kube-apiserver.

generator/transitive_closure.go
eloyekunle: I've added one

generator/transitive_closure.go
eloyekunle: Done

examples/import-boss/main.go
eloyekunle: It means we'll search for files up to the root of the filesystem.

E.g. Right now, if the program is run for `/home/user/go/src/bla.com/package`, `import-boss` will search for `.import-restrictions` up to `/home/user/go/src`, and stop there.

`$GOPATH`-less, if the program is run for `/home/user/code/my/package`, `import-boss` will search for `.import-restrictions` up to the `/home` directory.

examples/import-boss/generators/import_restrict.go
```diff
@@ -170,11 +182,14 @@ func removeLastDir(path string) (newPath, removedDir string) {
 }
 
 // Keep going up a directory until we find an .import-restrictions file.
```
sttts: doc what `next` does.

examples/import-boss/main.go
sttts: should we search up to a go.mod instead in non-GOPATH mode?

sttts:
Looks mostly good.

Can you squash in a sensible way?

examples/import-boss/main.go
eloyekunle: Yes, we can. Should we detect if Go modules are enabled or not, so maybe search up to `go.mod` if it is, and `$GOPATH/src` if it isn't?

sttts: would just look for the file and consider that dir as the root.

sttts: yes, $GOPATH is the fallback.

eloyekunle: Okay, sounds good.

eloyekunle: I've implemented the search up to `go.mod`, with $GOPATH as fallback.

examples/import-boss/generators/import_restrict.go
```diff
@@ -169,62 +181,104 @@ func removeLastDir(path string) (newPath, removedDir string) {
...
+			if err != nil {
+				return nil, err
+			}
+
+			restrictionFiles = append(restrictionFiles, rules)
 		}
 
 		nextPath, removedDir := removeLastDir(path)
-		if nextPath == path || removedDir == "src" {
+		if nextPath == path || isGoModRoot(path) || removedDir == "src" {
```
sttts: not related to this PR (follow-up): `src` is odd here, isn't it?

generator/default_package.go
```diff
@@ -27,6 +27,8 @@ type DefaultPackage struct {
 	// Import path of the package, and the location on disk of the package.
 	PackagePath string
 
+	Source string
```
sttts: go doc. What is source path? The current working directory?

examples/import-boss/generators/import_restrict.go
eloyekunle: That is the logic behind stopping the recursion at `$GOPATH/src`.

generator/default_package.go
eloyekunle: It's the location of the package on disk. I've added godoc.

examples/import-boss/generators/import_restrict.go
```diff
@@ -169,62 +182,104 @@ func removeLastDir(path string) (newPath, removedDir string) {
...
 	}
 
-	if rules == nil {
-		// No restrictions on this directory.
-		return nil
+	if err := irf.verifyRules(restrictionFiles, f); err != nil {
+		return err
+	}
+
+	if err := irf.verifyInverseRules(restrictionFiles, f); err != nil {
```
enxebre: nit: this could be inlined in the latest return

examples/import-boss/generators/import_restrict.go
```diff
@@ -169,62 +182,104 @@ func removeLastDir(path string) (newPath, removedDir string) {
...
 		path = nextPath
 	}
-	return nil, "", nil
+
+	return restrictionFiles, nil
 }
 
-func (importRuleFile) VerifyFile(f *generator.File, path string) error {
-	rules, actualPath, err := recursiveRead(path)
+func (irf importRuleFile) VerifyFile(f *generator.File, path string) error {
```
enxebre: `path` parameter seems to not be used within  `VerifyFile`

examples/import-boss/generators/import_restrict.go
```diff
@@ -169,62 +182,104 @@ func removeLastDir(path string) (newPath, removedDir string) {
 	return filepath.Join(filepath.Dir(dir), file), filepath.Base(dir)
 }
 
-// Keep going up a directory until we find an .import-restrictions file.
-func recursiveRead(path string) (*fileFormat, string, error) {
+// isGoModRoot checks if a directory is the root directory for a package
+// by checking for the existence of a 'go.mod' file in that directory.
+func isGoModRoot(path string) bool {
+	goModFile := filepath.Join(filepath.Dir(path), "go.mod")
```
enxebre: nit: this could be inlined, no need for the extra variable. Also may be put "go.mod" into a constant?

examples/import-boss/generators/import_restrict.go
```diff
@@ -243,8 +298,84 @@ func (importRuleFile) VerifyFile(f *generator.File, path string) error {
...
+}
+
+// verifyInverseRules checks that all packages that import a package are allowed to import it.
+func (irf importRuleFile) verifyInverseRules(restrictionFiles []*fileFormat, f *generator.File) error {
+	// compile all Selector regex in all restriction files
+	selectors := make([][]*regexp.Regexp, len(restrictionFiles))
+	for i, restrictionFile := range restrictionFiles {
+		for _, r := range restrictionFile.InverseRules {
+			re, err := regexp.Compile(r.SelectorRegexp)
+			selectors[i] = append(selectors[i], re)
```
enxebre: shall the `selectors[i]` assignment go after the error check?

examples/import-boss/generators/import_restrict.go
eloyekunle: Yeah, it's not used. But it fulfills an interface definition.

parser/parse.go
```diff
@@ -227,9 +227,16 @@ func (b *Builder) AddDirRecursive(dir string) error {
...
+	// filepath.Walk.
+	realPath, err := filepath.EvalSymlinks(b.buildPackages[dir].Dir)
+	if err != nil {
+		return err
+	}
+
 	// filepath.Walk includes the root dir, but we already did that, so we'll
 	// remove that prefix and rebuild a package import path.
-	prefix := b.buildPackages[dir].Dir
+	prefix := realPath
```
enxebre: I see this was also in the original code but what's the purpose of this variable?

eloyekunle: It's used in the following block of code, but I've removed it and used `realPath` there.

PR #155, opened 2019-10-01
Adding a new generic, re-usable conversion-gen example

wk8:
This patch is taking conversion-gen out of the code-generator repo, and instead adding it here as a re-usable, generic example; for the sake of being able to re-use the conversion logic in other generators - this will be re-used immediately in [csi-proxy](https://github.com/kubernetes-csi/csi-proxy/tree/15613db6b0fa6b2b183e1ecc52b3bfd64b4a5005/client/api#detailed-breakdown-of-generated-files).

A new package, `k8s.io/gengo/examples/conversion-gen/generators/generator`, exposes a new `ConversionGenerator` public struct that fulfills the `Generator` interface, and can be re-used in other repos. That structure is wrapped in a private `genConversion` struct that adds the logic specific to kubernetes.

This patch could be adopted as-in in k8s itself, as a drop-in replacement for the current conversion-gen code, as demonstrated by https://github.com/kubernetes/kubernetes/pull/83345.

Added a few tests.

Signed-off-by: Jean Rouge <rougej+github@gmail.com>


examples/conversion-gen/generators/conversion.go
```diff
@@ -0,0 +1,339 @@
...
+	conversionargs "k8s.io/code-generator/cmd/conversion-gen/args"
+)
+
+// These are the comment tags that carry parameters for conversion generation.
+const (
+	// e.g., "+k8s:conversion-gen=<peer-pkg>" in doc.go, where <peer-pkg> is the
+	// import path of the package the peer types are defined in.
+	// e.g., "+k8s:conversion-gen=false" in a type's comment will let
+	// conversion-gen skip that type.
+	tagName = "k8s:conversion-gen"
```
lavalamp: No kubernetes specific stuff should go in this repo.

examples/conversion-gen/Makefile
```diff
@@ -0,0 +1,16 @@
+TOOL=conversion-gen
```
lavalamp: blargh, when did this repo grow makefiles :( :(

lavalamp:
Why does this need to go here? Why not make changes to the code generator directly?

examples/conversion-gen/generators/conversion.go
wk8: Eh, I would tend to agree, but then I followed the pattern set by e.g. deepcopy or defaultergen, where there are a lot of k8s-specific stuff...

wk8:
> Why does this need to go here? Why not make changes to the code generator directly?

As mentioned in the commit message, the goal here is to have a generic, re-usable conversion generator, not just for k8s; code-generator is specifically meant for k8s-style APIs, which I think is narrower than what we want here.

examples/conversion-gen/generators/conversion.go
lavalamp: Yeah, sigh, I am not sure how those got in here with k8s specific tags.

examples/conversion-gen/Makefile
sttts: long long ago

examples/conversion-gen/generators/conversion.go
sttts: Where is deepcopy kubernetes-dependent? Some test cases use Kubernetes strings. But generation is generic.

lavalamp: here: https://github.com/kubernetes/gengo/blob/master/examples/deepcopy-gen/generators/deepcopy.go#L43

wk8: @sttts : it's the same situation here, pretty much.

sttts: ok, that's a string. That does not count :)

Doesn't conversion-gen use types like Scheme?

wk8: Only for the example.

If that's a problem though, happy to make the example simpler, and move that logic back to code-generator.

sttts:
what is the reason to move conversion-gen here? Because k8s.io/code-generator has unpleasant dependencies? Why not move deepcopy-gen and default-gen to k8s.io/code-generator? Or to create sigs.k8s.io/generic-code-generation or something like that?

wk8:
@sttts : the main goal of this PR is to have a re-usable and extensible conversion generator; that's really what I care about here - to be able to re-use that generator in https://github.com/kubernetes-csi/csi-proxy

The accepted pattern so far seemed to be to have generic generators as examples in this repo, and besides code-generator claims to be home to "generators for kube-like API types", which sounds a little narrower than what we're aiming for here.

Now if we all agree that moving all examples to code-generator (and then broadening the scope of that repo), or creating a new sigs.k8s.io/generic-code-generation lib is a better way to do this, sounds good to me :)

sttts:
> re-usable and extensible conversion generator

Have not read you PR in detail. What have you changed here to make it more generic than the code-generator one?

> Now if we all agree that moving all examples to code-generator (and then broadening the scope of that repo), or creating a new sigs.k8s.io/generic-code-generation lib is a better way to do this, sounds good to me :)

I think the example here in the repo developed form simple proof-of-concept examples to generators used for production code in kube. I can only speak about deepcopy-gen, but that one is complex enough to go elsewhere, k8s.io/code-generator would be a better place actually.

wk8:
> Have not read you PR in detail. What have you changed here to make it more generic than the code-generator one?

I took the k8s-specific logic out into conversion.go (https://github.com/kubernetes/gengo/pull/155/files#diff-c03c0831605f88a11edfd53e5d434588) - e.g. the way k8s handles vendored deps, its own custom args with base peer dirs, how it deals with converting structs from one package to the other (ie with `conversion.Scope`s, how it deals with missing fields in peer types, the init function it needs for conversion schemes, etc...
And separated that from the generic task of looking at 2 structs with the same name in 2 different packages, and generating code from one to the other; that's now encapsulated in the `ConversionGenerator` struct (https://github.com/kubernetes/gengo/pull/155/files#diff-98d9c6af98d89c316c729f1c059379f2R30) and be re-used elsewhere, and configured as needed.

> I think the example here in the repo developed form simple proof-of-concept examples to generators used for production code in kube.

Agreed.

> I can only speak about deepcopy-gen, but that one is complex enough to go elsewhere, k8s.io/code-generator would be a better place actually.

Happy to move all the complex, actually used as libs in k8s, examples to code-generator if you think that's a better home for them. @lavalamp : what do you make of that idea?

sttts:
> I took the k8s-specific logic out into conversion.go (https://github.com/kubernetes/gengo/pull/155/files#diff-c03c0831605f88a11edfd53e5d434588) - e.g. the way k8s handles vendored deps, its own custom args with base peer dirs, how it deals with converting structs from one package to the other (ie with `conversion.Scope`s, how it deals with missing fields in peer types, the init function it needs for conversion schemes, etc...
> And separated that from the generic task of looking at 2 structs with the same name in 2 different packages, and generating code from one to the other; that's now encapsulated in the `ConversionGenerator` struct (https://github.com/kubernetes/gengo/pull/155/files#diff-98d9c6af98d89c316c729f1c059379f2R30) and be re-used elsewhere, and configured as needed.

From that this sounds like to be enough code to justify its very own repository or a package in your project. What concerns me most here is the maintenance. Who will update it if gengo changes? We cannot collect generic generators in here and magically maintain them (our resources are already very limited).

If I don't misunderstand you, your code is essentially an incompatible fork of a non-trivial generator that cannot replace the one used today in kube. Is that the right understanding? In that case I tend to suggest to separate it from gengo and code-generator into github.com/xyz/conversion-gen or so.

wk8:
> If I don't misunderstand you, your code is essentially an incompatible fork of a non-trivial generator that cannot replace the one used today in kube.

I'm sorry, I did a terrible job explaining :) it's not incompatible at all. Quite the opposite. It separates the k8s-specific logic from the generic conversion logic, then changes k8s' generator to use the new, generic generator, configuring it to do what k8s needs. The resulting generator is a drop-in replacement for the existing k8s generator, as demonstrated by https://github.com/kubernetes/kubernetes/pull/83345 - that PR is certainly not meant to be merged as is, but it does show that the new k8s generator, using the generic generator under the hood, produces the exact same code as the old one.

 

> From that this sounds like to be enough code to justify its very own repository or a package in your project. What concerns me most here is the maintenance.

Precisely, the goal here is to have the generic conversion logic in a single place, so that all repos using it benefit from maintenance and improvements to it. So sure, it could be a package in csi-proxy, but then we'd have to maintain it there, as well as the existing conversion logic in k8s; seems better to have it all in a central place.

sttts:
Thanks for the explanation, this changes the discussion :)

Would it make sense to also keep the generic generator in k8s.io/code-generator and put a kube-specific one just next to it? like `k/code-generator/cmd/generic-conversion-gen`? I think that's most pragmatic and avoid moving more production-code generators here into examples.

wk8:
> Would it make sense to also keep the generic generator in k8s.io/code-generator and put a kube-specific one just next to it? like `k/code-generator/cmd/generic-conversion-gen`? I think that's most pragmatic and avoid moving more production-code generators here into examples.

Absolutely, that's a great idea. If @lavalamp is on board with that, happy to do it ASAP.

wk8:
As discussed on Slack with @sttts , moving back to code-generator, and then have a generic-conversion-gen, and a k8s-conversion-gen, both using the same package under the hood for the generic parts.

PR #160, opened 2020-01-01
Use UTC for calculation of year

johngmyers:
/kind bug

In the evening/night of December 31 in the Western hemisphere, gengo will create output that fails CI validation.

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -112,7 +112,7 @@ func (g *GeneratorArgs) LoadGoBoilerplate() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	b = bytes.Replace(b, []byte("YEAR"), []byte(strconv.Itoa(time.Now().Year())), -1)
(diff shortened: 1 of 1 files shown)


args/args.go
```diff
@@ -112,7 +112,7 @@ func (g *GeneratorArgs) LoadGoBoilerplate() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	b = bytes.Replace(b, []byte("YEAR"), []byte(strconv.Itoa(time.Now().Year())), -1)
+	b = bytes.Replace(b, []byte("YEAR"), []byte(strconv.Itoa(time.Now().UTC().Year())), -1)
```
lavalamp: I guess the verification script / server is using UTC? This makes sense from that perspective, but it will be confusing for a user expecting their tools to use their local time zone.

johngmyers: Travis does (as should all cloud servers); don't know about Prow. But as long as CI uses this same code, it will agree on which year should be in comments.

PR #161, opened 2020-01-14
Use YAML format for import-restrictions

eloyekunle:
As suggested by @sttts, this will enable us add comments to import-restrictions files giving explanations for some import restrictions that we add.


eloyekunle:
checking case sensitivity
/hold

PR #165, opened 2020-02-01
reduce noise in import-boss log output

eloyekunle:
Running import-boss on Kubernetes produces hundreds of instances of logs like below which increases noise in log output:
`
W0117 19:40:19.480385  296076 parse.go:246] Ignoring child directory k8s.io/apiserver/pkg/server/options/encryptionconfig/testdata/valid-configs/kms: unable to import "k8s.io/apiserver/pkg/server/options/encryptionconfig/testdata/valid-configs/kms": cannot find package "k8s.io/apiserver/pkg/server/options/encryptionconfig/testdata/valid-configs/kms" in any of:
	/home/prow/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/server/options/encryptionconfig/testdata/valid-configs/kms (vendor tree)
	/usr/local/go/src/k8s.io/apiserver/pkg/server/options/encryptionconfig/testdata/valid-configs/kms (from $GOROOT)
	/home/prow/go/src/k8s.io/kubernetes/_output/local/go/src/k8s.io/apiserver/pkg/server/options/encryptionconfig/testdata/valid-configs/kms (from $GOPATH)
`

This specific error occurs when there is no go source in a package and subpackages.

This PR logs the error with higher verbosity which reduces log ouput noise.

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -28,6 +28,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -334,6 +335,12 @@ func (b *Builder) addDir(dir string, userRequested bool) error {
 	return nil
 }
 
+var regexErrPackageNotFound = regexp.MustCompile(`^unable to import ".*?": cannot find package ".*?" in any of:`)
+
+func isErrPackageNotFound(err error) bool {
+	return regexErrPackageNotFound.MatchString(err.Error())
+}
+
 // importPackage is a function that will be called by the type check package when it
 // needs to import a go package. 'path' is the import path.
 func (b *Builder) importPackage(dir string, userRequested bool) (*tc.Package, error) {
@@ -356,6 +363,11 @@ func (b *Builder) importPackage(dir string, userRequested bool) (*tc.Package, er
 
 		// Add it.
 		if err := b.addDir(dir, userRequested); err != nil {
+			if isErrPackageNotFound(err) {
+				klog.V(6).Info(err)
+				return nil, nil
+			}
+
 			return nil, err
 		}
 
--- a/parser/local_parse_test.go
+++ b/parser/local_parse_test.go
@@ -36,6 +36,13 @@ func TestImportBuildPackage(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


parser/parse.go
```diff
@@ -127,7 +136,7 @@ func (b *Builder) importBuildPackage(dir string) (*build.Package, error) {
 	buildPkg, err := b.importWithMode(dir, build.ImportComment)
 	if err != nil {
 		if _, ok := err.(*build.NoGoError); !ok {
-			return nil, fmt.Errorf("unable to import %q: %v", dir, err)
+			return nil, &importError{dir: dir, err: err}
```
sttts: would rather follow https://blog.golang.org/go1.13-errors with error wrapping with `%w`.

Note that the type cast below would not work if you return a pointer here.

eloyekunle: The specific error we're looking for isn't a concrete type: https://github.com/golang/go/blob/master/src/go/build/build.go#L739 so wrapping won't work unless we wrap our custom type.

sttts: Couldn't you check for `*build.NoGoError`?

sttts: Oh, `*build.NoGoError` is what you want log output for? Am confused.

eloyekunle: `*build.NoGoError` is [already ignored](https://github.com/kubernetes/gengo/blob/ae764952c2a6af9c1b90ab38c9564b29623962ad/parser/parse.go#L138) in the `importBuildPackage` function so it's not returned.

parser/parse.go
```diff
@@ -116,6 +116,11 @@ func (b *Builder) AddBuildTags(tags ...string) {
 	b.context.BuildTags = append(b.context.BuildTags, tags...)
 }
 
+func isErrPackageNotFound(err error) bool {
+	errMessage := err.Error()
+	return strings.HasPrefix(errMessage, "cannot find package") && strings.Contains(errMessage, "in any of:")
```
sttts: is this testable in some lines of code that the strings do not change in another go release?

sttts:
Would increase the verbosity even more, to 6 IMO. Then lgtm.

PR #167, opened 2020-02-03
Convert path to slashes before passing to build.Context.Import()

### LINKED ISSUE #166: Gengo passes file path as package path to build.Context.Import().
In parse.go, the function `Builder.importWithMode()` calls `build.Context.Import()` with `dir` as the first argument. But `dir` is a _filesystem path_, as opposed to a package path, and `Import` expects _a package path_ as its first argument.

On UNIX systems, this doesn't matter because the two turn out to be the same thing. On Windows, it is a huge problem because package paths are _always_ slash separated while filesystem paths use backslashes.

al45tair:
On UNIX systems, we get away with the existing code because filesystem
and package paths both use the same kind of separator.  On Windows,
sadly, that isn't the case and that causes everything to go wrong
(more specifically, we end up not being able to detect that this is a
local package, and so the output path for the generated files is
wrong).

Fixes issue #166.

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -589,7 +589,7 @@ func (b *Builder) importWithMode(dir string, mode build.ImportMode) (*build.Pack
 	if err != nil {
 		return nil, fmt.Errorf("unable to get current directory: %v", err)
 	}
-	buildPkg, err := b.context.Import(dir, cwd, mode)
+	buildPkg, err := b.context.Import(filepath.ToSlash(dir), cwd, mode)
 	if err != nil {
 		return nil, err
 	}


lavalamp:
Hm, I think this code is part of trying to figure out what the right import path should be for the given directory. So this seems... hacky, but the code here is already pretty questionable, so ok I guess?

(but you need to sign the CLA)

al45tair:
@lavalamp @sttts CLA has been signed. Any chance this can go in?

sttts:
How can we test this continuously?

lavalamp:
I filed https://github.com/kubernetes/gengo/issues/174. I definitely expect regressions for things that aren't tested. But I think blocking a 1 line change for lack of a big testing setup is maybe overkill.

/lgtm
/approve

PR #169, opened 2020-02-06
Fix pointer dereferencing in example generator

ubombi:
Current code generates 
```go
func (in *TestType) DeepCopyObject() runtime.Object {
	if c := in.DeepCopy(); c != nil {
		return c
	}
	return nil
}
```
Which makes no much sense and should be simplified to  `return in.DeepCopy()`
On the other hand, there is no `nil` check for nonpointer-interfaces.

```go
func (in TestType) DeepCopyObject() runtime.Object {
	return *in.DeepCopy();
}
```
which is not a big deal as `TestType` is not a pointer receiver. But if user defines it's own `.DeepCopy()` it can be a good idea to check that pointer.

After this patch, you get 
```go
func (in *TestType) DeepCopyObject() runtime.Object {
	return in.DeepCopy()
}
```
and 
```go
func (in TestType) DeepCopyObject() runtime.Object {
	if c:= in.DeepCopy(); c != nil {
		return *c
	}
	return nil
}
```


lavalamp:
Can you add or adjust one of the test cases in https://github.com/kubernetes/gengo/tree/master/examples/deepcopy-gen/output_tests?

ubombi:
Sure, will do

sttts:
> ```go
> ```go
> func (in *TestType) DeepCopyObject() runtime.Object {
> 	if c := in.DeepCopy(); c != nil {
> 		return c
> 	}
> 	return nil
> }
> ```
> 
> Which makes no much sense and should be simplified to  return in.DeepCopy()

This conclusion is wrong. Golang makes a difference between nil interfaces and non-nil interfaces will nil value. This cannot be optimized like that.

/hold

sttts:
> which is not a big deal as `TestType` is not a pointer receiver. But if user defines it's own `.DeepCopy()` it can be a good idea to check that pointer.

Why? A `DeepCopy` on a non-nil reference should give a non-nil object. Otherwise, the custom code does not implement `DeepCopy` correctly.

ubombi:
> nil interfaces and non-nil interfaces will nil value
1. My bad. Forgot about that. 

> the custom code does not implement DeepCopy correctly
2. And still, I don't think this is a reason to `panic`.

PR #170, opened 2020-03-16
Record and store parameters name to function's signature.

longit644:
Record and store parameters name to function's signature.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -413,10 +413,10 @@ func (m Member) String() string {
 // Signature is a function's signature.
 type Signature struct {
 	// If a method of some type, this is the type it's a member of.
-	Receiver           *Type
-	Parameters         []*Type
-	ParametersWithName []Parameter
-	Results            []*Type
+	Receiver       *Type
+	Parameters     []*Type
+	ParameterNames []string
+	Results        []*Type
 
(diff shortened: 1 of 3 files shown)


types/types.go
```diff
@@ -412,12 +412,11 @@ func (m Member) String() string {
...
 type Signature struct {
-	// TODO: store the parameter names, not just types.
-
 	// If a method of some type, this is the type it's a member of.
-	Receiver   *Type
-	Parameters []*Type
-	Results    []*Type
+	Receiver           *Type
+	Parameters         []*Type
+	ParametersWithName []Parameter
```
lavalamp: `NamedParameters`

What do you think about instead just doing `ParameterNames []string`? It is awkward (I hate parallel arrays), but doesn't duplicate information or leave people scratching their heads quite as much about why we made the interface what it is.

lavalamp: We can claim they're separate since if the parameters are unnamed, they must all be unnamed :)

longit644: Yes, It makes sense. Using `ParameterNames []string` isn't the best practice, but it makes the interface clear.

longit644:
Hi all, do I need to update anything?

PR #172, opened 2020-04-09
add int8 and uint8 builtin types

### LINKED ISSUE #171: Int8 should not be byte type
Current builtin types in gengo

```
		Types: map[string]*Type{
			"bool":    Bool,
			"string":  String,
			"int":     Int,
			"int64":   Int64,
			"int32":   Int32,
			"int16":   Int16,
			"int8":    Byte,
			"uint":    Uint,
			"uint64":  Uint64,
			"uint32":  Uint32,
			"uint16":  Uint16,
			"uint8":   Byte,
			"uintptr": Uintptr,
			"byte":    Byte,
			"float":   Float,
			"float64": Float64,
			"float32": Float32,
		}
```

### Description

int8 and uint8 are basic kind in go and defined as reflect.Kind (See https://golang.org/pkg/reflect/#Kind)

However byte which is uint8 kind actually is not basic kind

### Expected

```
int8 is Int8 type, uint8 is Uint8 type and byte is Uint8 type
```

### Actual

```
Now int8, uint8 and byte are Byte type
```

liubog2008:
Fix #171

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -445,6 +445,10 @@ var (
 		Name: Name{Name: "int16"},
 		Kind: Builtin,
 	}
+	Int8 = &Type{
+		Name: Name{Name: "int8"},
+		Kind: Builtin,
+	}
 	Int = &Type{
 		Name: Name{Name: "int"},
 		Kind: Builtin,
@@ -459,6 +463,10 @@ var (
 	}
 	Uint16 = &Type{
 		Name: Name{Name: "uint16"},
+		Kind: Builtin,
+	}
+	Uint8 = &Type{
+		Name: Name{Name: "uint8"},
 		Kind: Builtin,
 	}
 	Uint = &Type{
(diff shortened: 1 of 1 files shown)


lavalamp:
Is there really no test impact from this?

I vaguely recall originally having made them all byte since I thought there was something (I don't remember what) that wouldn't be able to handle different names for the "same" type.

liubog2008:
I have searched `types.Byte` in code-generator.

Only the below code uses it.

https://github.com/kubernetes/code-generator/blob/master/cmd/conversion-gen/generators/conversion.go#L1159

liubog2008:
> I vaguely recall originally having made them all byte since I thought there was something (I don't remember what) that wouldn't be able to handle different names for the "same" type.

lavalamp:
It looks like this fell off my radar earlier, apologies...

I think I'd be happy to fix this, but I'm nervous about merging it if we haven't verified that vendoring the change into the main kubernetes repo doesn't break anything. The search is encouraging but I'd like to be sure. If you are still interested in this, and have time to try that and report back, that'd be sufficient for me, I think.

longit644:
Hi all, any update on this? Thanks.

PR #175, opened 2020-04-25
fix method with same sigature comments mismatch

zexi:
Fix:
Different methods with same signature will get same commentLines.


lavalamp:
/lgtm
/approve

thanks!

PR #178, opened 2020-05-05
Update gengo to use go mod.

n3wscott:
As a replacement for https://github.com/lavalamp/gengo/pull/1 and #144.

Fixes https://github.com/kubernetes/gengo/issues/143

The move to go 1.13 seems to have solved the previous issues found in go mod 1.12. 

Signed-off-by: Scott Nichols <snichols@vmware.com>


n3wscott:
/retest

not sure how to get travis to run again.

n3wscott:
Looks ready for test (if there are prow tests), travis passed now.

hack/update-deps.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
lavalamp, line 1: I'd love to keep this repo 100% shell-free :/

Is there any alternative to adding this? @liggitt might have an idea?

go.mod
```diff
@@ -0,0 +1,20 @@
...
+	github.com/google/go-cmp v0.4.0
+	github.com/google/gofuzz v1.1.0
+	github.com/spf13/pflag v1.0.5
+	golang.org/x/tools v0.0.0-20200505023115-26f46d2f7ef8
+	k8s.io/klog/v2 v2.0.0
+	sigs.k8s.io/yaml v1.2.0
+)
+
+replace (
+	fake v0.0.0 => ./fake // vendor the fake dep
```
liggitt, line 17: this is definitely unexpected. if we have modules as test dependencies put them under a testdata folder or something similar

go.mod
```diff
@@ -0,0 +1,20 @@
+module k8s.io/gengo
+
+go 1.13
+
+require (
+	fake v0.0.0
```
liggitt, line 6: drop this, it will be inherited by all dependants

hack/tools.go
```diff
@@ -0,0 +1,21 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hack
+
+// Imports to ensure vendoring with go mod.
+
+import _ "fake/dep" // fake/dep is used for testing.
```
liggitt, line 21: drop this… see comment about using testdata or something similar

hack/update-deps.sh
liggitt: I like the approach of super-minimal makefiles (e.g. one-line target definitions, no external file includes) that just call go commands, e.g.:

make test == go test ./…
make deps == go mod tidy && go mod vendor

It technically is still shell-adjacent, but with those two rules it tends to be less of a dumping ground for crazy stuff.

Also, is it necessary to vendor?

n3wscott: Not needed to vendor but I think it was originally, I can double check.

go.mod
n3wscott: testdata is not going to work with go mod from what I can tell. The original testing pattern is faking out go mod, or the testdata folder needs a go.mod file. I can try that.

go.mod
n3wscott: 😬 right, ok.

go.mod
liggitt: you would need to set the current working dir to the folder under testdata which defined the go.mod file

hack/update-deps.sh
n3wscott: a fake vendor directory does not work in go 1.14, it confuses go mod to only look in vendor and not the cache. I moved the fake package to testdata.

go.mod
n3wscott: I got it to work! but go mod added an indirect import, and I added a replace directive to allow it to find the sub-module. 

This should get trimmed off when something deps on this package.

go.mod
n3wscott: Done.

hack/tools.go
n3wscott: Dropped.

liggitt:
> I got it to work! I added a replace directive to allow it to find the sub-module.
>
> This should get trimmed off when something deps on this package.

That still seems like it shouldn't be necessary. I'd expect test data to be completely isolated from the dependencies of the module.

See https://github.com/kubernetes-sigs/clientgofix/blob/v0.3.0/pkg/fix_test.go#L38-L42 for an example of exercising various $GOPATH/module vendor/no-vendor scenarios in test cases

n3wscott:
> > I got it to work! I added a replace directive to allow it to find the sub-module.
> > This should get trimmed off when something deps on this package.
> 
> That still seems like it shouldn't be necessary. I'd expect test data to be completely isolated from the dependencies of the module.
> 
> See https://github.com/kubernetes-sigs/clientgofix/blob/v0.3.0/pkg/fix_test.go#L38-L42 for an example of exercising various $GOPATH/module vendor/no-vendor scenarios in test cases

but the point of this test is that there is a package that is out of tree that the test is solving for, so go mod needs to know where to look. Previously it looked in vendor as that is a magic extension of $GOPATH, but that option is not available for go 1.14 as the existence of a vendor dir will limit the $GOPATH to only `./vendor` by default. The replace directly allows go mod to locate the package in `./testdata`.

n3wscott:
Here is the test that uses the package `fake/dep` https://github.com/kubernetes/gengo/blob/e0e292d8aa122d458174e1bef5f142b4d0a67a05/parser/local_parse_test.go#L25

go mod also does not see this as an import because package `fake/dep` is not used in the go code, only this test call. So this means it only needs to replace to point to a local go.mod file.

Without the replace line, the test fails:

```=== RUN   TestImportBuildPackage
    TestImportBuildPackage: local_parse_test.go:26: unable to import "fake/dep": package fake/dep is not in GOROOT (/usr/local/Cellar/go/1.14.2_1/libexec/src/fake/dep)
--- FAIL: TestImportBuildPackage (0.04s)
```

This is expected because of the vendor gotchyas with go 1.14

n3wscott:
Ping on this. RFAL

lavalamp:
I'm OK with this. `fake` is only in the replace directives, not in the e.g. `go.sum` file. Not vendoring makes this much more palatable :)

Jordan, any more thoughts?

liggitt:
I really don't want fake test modules or dependencies expressed in the go.mod file. I added a commit to https://github.com/liggitt/kubernetes-gengo/commits/move-go-mod which lets us run this test scenario without a replace in the actual go.mod

lavalamp:
Yeah, that looks better.

n3wscott:
ok thanks @liggitt I merged your update into this PR to drop the replace.

lavalamp:
/lgtm
/approve

Thanks!!

PR #179, opened 2020-05-20
fix tcNameToName

u2takey:
if type is method, format of `in` will be `func (a x) method(){}`, missing space before `(`

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -628,7 +628,7 @@ func tcNameToName(in string) types.Name {
 		strings.HasPrefix(in, "<-chan") ||
 		strings.HasPrefix(in, "chan<-") ||
 		strings.HasPrefix(in, "chan ") ||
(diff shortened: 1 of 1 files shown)


lavalamp:
Can you add a test, preferably one that uses the go parser libraries to generate these names, so that we'll be able to detect if this changes?

u2takey:
> Can you add a test, preferably one that uses the go parser libraries to generate these names, so that we'll be able to detect if this changes?

i found some `Name` with weird package and name, for example (`tcNameToName` https://github.com/kubernetes/gengo/blob/master/parser/parse.go#L624  Output) :

```
in = func (reflect.Type).AssignableTo(u reflect.Type) bool
name.Pacakge = func (reflect.Type).AssignableTo(u reflect
name.Name = Type) bool
```
But this pr haven't fix all these weird cases after some test, so closing it for now.

PR #185, opened 2020-09-29
Add support for generating defaults for external types

detiber:
Adds support for generating defaults for external types, similar to what conversion-gen does.

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -41,6 +41,7 @@ type CustomArgs struct {
 // These are the comment tags that carry parameters for defaulter generation.
 const tagName = "k8s:defaulter-gen"
 const inputTagName = "k8s:defaulter-gen-input"
+const externalTypesTagName = "k8s:defaulter-gen-external-types"
 
 func extractTag(comments []string) []string {
 	return types.ExtractCommentTags("+", comments)[tagName]
@@ -48,6 +49,10 @@ func extractTag(comments []string) []string {
 
 func extractInputTag(comments []string) []string {
 	return types.ExtractCommentTags("+", comments)[inputTagName]
+}
+
+func extractExternalTypesTag(comments []string) []string {
+	return types.ExtractCommentTags("+", comments)[externalTypesTagName]
 }
 
 func checkTag(comments []string, require ...string) bool {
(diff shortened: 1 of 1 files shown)


detiber:
@sttts I ran into this issue when trying to add Namespaces support to my minimal apiserver implementation, I used the changes here to generate the defaults used here: https://github.com/thetirefire/badidea/pull/12/files#diff-a80ecef7b7a6df7985c380dd10060f26

sttts:
Why would you want defaults for external types?

detiber:
@sttts In the current case I'm trying to re-implement support for namespaces in https://github.com/thetirefire/badidea/pull/12, but we've also hit similar types of issues with Cluster API when trying to expose kubeadm config types. 

Since the defaults (and validations) apply to the public types, it makes sense to either be able to import and/or generate them for various reasons in external projects.

sttts:
I am not convinced this is needed.

Please add an example (as simple as possible) to the `_output_tests` directory.

/hold

lavalamp:
Whoops, sorry, missed this until now-- @apelisse and @sttts have been working on defaults.

detiber:
@lavalamp I'm not sure the defaulting changes (at least the ones in https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1929-built-in-default and the related changes in https://github.com/kubernetes/enhancements/pull/2082/files) would resolve the issue.

Basically, since the defaulting logic is not defined as part of k8s.io/api, if I'm trying to re-use those types fully (not embedding within another type or CRD), there is no current way for me to use the same defaulting as k/k (without also importing k/k).

Currently the generated defaults in k/k are generated using hardcoded paths to the external types defined in the staging repository. While I suppose it would be possible to also introduce a similar staging directory and generate the defaulting logic the same way it is done in k/k, however that also comes with all the same downsides.

By allowing defaulter-gen to reference external types, it would also benefit a potential future k/k that leverages go modules without a staging directory.

Alternatively, I would love it if the defaulting logic was in k8s.io/api, since that would also allow for easier use in other places (such as client side code), but that does present more complications for updating or changing defaults.

sttts:
@detiber you have k/api vendored. Why not just reference the vendored copy?

> By allowing defaulter-gen to reference external types, it would also benefit a potential future k/k that leverages go modules without a staging directory.

I don't follow here. We do exactly that:

k/pkg/apis/apps/v1/doc.go:
```
// +k8s:defaulter-gen-input=../../../../vendor/k8s.io/api/apps/v1
```

It references external types. What is different in your use-case (other than that `../../../../vendor/k8s.io/api` is a real directory and not a symlink) ?

sttts:
> Why would you want defaults for external types?

To clarify: defaulting is always operating on external types. But defaulting is something we don't want non-apiservers to consume. Hence, we have it in `k/pkg/apis/<group>/<version>. The "real" external types package is at an external place already today (via `+k8s:defaulter-gen-input`).

PR #188, opened 2020-10-19
Capture the values of constant declarations.

### LINKED ISSUE #187: Capture the underlying value of constants.
When using gengo to generate API documentation, it would be pretty convenient to be able to get the underlying [value](https://godoc.org/go/types#Const.Val) of the constant. This could be used to generate documentation for the acceptable values of an enum.

jpeach:
Capture the value of constants declarations in `parse.Type`. This allows
client code to inspect the values of all constants in a package using
the package's `.Constants` map.

This fixes #187.

### DIFF
--- a/parser/parse_test.go
+++ b/parser/parse_test.go
@@ -24,7 +24,9 @@ import (
 	"testing"
 	"text/template"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"k8s.io/gengo/args"
 	"k8s.io/gengo/namer"
@@ -157,10 +159,24 @@ func TestBuilder(t *testing.T) {
 	                AnotherVar = Frobber{}
                 )
 
-				type Enumeration string
-				const (
-					EnumSymbol Enumeration = "enumSymbol"
-				)
+		type Enumeration string
+		const (
+			EnumSymbol Enumeration = "enumSymbolValue"
+		)
+
+		type Degrees int
+		const (
+			FirstDegree Degrees = iota
+			SecondDegree
+			ThirdDegree
+		)
+
+		const ConstNineNine = 99
+		const ConstHundred = ConstNineNine + 1
+
+		const ConstExpr = 1 - 0.707i * 9.3
+		const ConstFloat = float64(7.8)
+		const ConstString = "constant string"
                 `,
 		},
 	}
@@ -179,7 +195,7 @@ package o
 {{define "Var"}}{{$t := .Underlying}}var {{Name .}} {{Raw $t}} = {{Raw .}}
 
 {{end}}
-{{define "Const"}}{{$t := .Underlying}}const {{Name .}} {{Raw $t}} = {{Raw .}}
+{{define "Const"}}{{$t := .Underlying}}const {{Name .}} {{Raw $t}} = {{Raw .}}({{ .ConstValue }})
 
 {{end}}
 {{range $t := .}}{{if eq $t.Kind "Struct"}}{{template "Struct" $t}}{{end}}{{end}}
@@ -228,7 +244,13 @@ var FooAVar proto.Frobber = proto.AVar
 var FooAnotherVar proto.Frobber = proto.AnotherVar
 
 
-const FooEnumSymbol proto.Enumeration = proto.EnumSymbol
+const FooEnumSymbol proto.Enumeration = proto.EnumSymbol(enumSymbolValue)
+
+const FooFirstDegree proto.Degrees = proto.FirstDegree(0)
+
+const FooSecondDegree proto.Degrees = proto.SecondDegree(1)
+
+const FooThirdDegree proto.Degrees = proto.ThirdDegree(2)
 
 `
 	testNamer := namer.NewPublicNamer(1, "proto")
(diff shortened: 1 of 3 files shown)


parser/parse_test.go
```diff
@@ -253,6 +263,18 @@ const FooEnumSymbol proto.Enumeration = proto.EnumSymbol
...
+	}
+	for k, v := range map[string]string{"EnumSymbol": `"enumSymbolValue"`, "TenBytes": "10", "ConstNumber": "99"} {
+		if c, ok := u.Package("base/foo/proto").Constants[k]; !ok {
+			t.Errorf("Missing constant %q", k)
+		} else {
+			if c.Constval == nil || *c.Constval != v {
+				t.Errorf("Unexpected value %#v for constant %q", v, k)
+			}
+		}
+	}
```
nikhita, line 277: optional nit: this could probably be simpler with `cmp.Diff(expectedConstants, u.Package("base/foo/proto").Constants))`....although `expectedConstants` would need to be `map[string]*types.Type` instead of `map[string]string`

nikhita (commented):
:+1: 

/assign @sttts

parser/parse_test.go
jpeach: Good point; fixed!

jpeach:
ping @sttts @lavalamp

parser/parse_test.go
```diff
@@ -159,8 +161,15 @@ func TestBuilder(t *testing.T) {
 
 				type Enumeration string
 				const (
-					EnumSymbol Enumeration = "enumSymbol"
+					EnumSymbol Enumeration = "enumSymbolValue"
 				)
+
+				type Bytes int
```
lavalamp, line 167: I know it's test code, but "Bytes int" is confusing :)

parser/parse_test.go
```diff
@@ -159,8 +161,15 @@ func TestBuilder(t *testing.T) {
...
+				const (
+					TenBytes Bytes = 10
+				)
+
```
lavalamp, line 171: can we test some more complicated cases like `const foo = []byte("foo")` or `const bar = float64(1)`?

types/types.go
```diff
@@ -341,6 +341,11 @@ type Type struct {
 	// If Kind == func, this is the signature of the function.
 	Signature *Signature
 
+	// Constval contains a stringified constant value if
+	// Kind == DeclarationOf and this is a constant value
+	// declaration.
+	Constval *string
```
lavalamp, line 347: Can we go with `ConstValue` for the name?

parser/parse_test.go
lavalamp: I'm slightly concerned that this feature will be hard to use when it's a cast, I don't know what these will print.

Also what about:

```
const foo = 1
const bar = foo+1

const (
  x = iota
  y
  z
)
```

types/types.go
jpeach: Fixed.

parser/parse_test.go
jpeach: Renamed to "Degrees"

parser/parse_test.go
jpeach: The `[]byte("foo")` isn't a legal Go constant, so no need to worry about that case. The simple addition and `iota` case work as you probably expect, which is great.

The const [Value](https://golang.org/pkg/go/constant/#Value) for the floating point values surprised me a lot and is different between String and ExactString (see diff below which is from switching to String):

```
    parse_test.go:330: Constant mismatch:   map[string]*types.Type{
          	"ConstExpr": &{
          		... // 1 ignored and 7 identical fields
          		Methods:    nil,
          		Signature:  nil,
        - 		ConstValue: &"(1 + -6.5751i)",
        + 		ConstValue: &"(1 + -65751/10000i)",
          	},
          	"ConstFloat": &{
          		... // 1 ignored and 7 identical fields
          		Methods:    nil,
          		Signature:  nil,
        - 		ConstValue: &"7.8",
        + 		ConstValue: &"8782019273372467/1125899906842624",
          	},
          	"ConstHundred":  &{Name: types.Name{Package: "base/foo/proto", Name: "ConstHundred"}, Kind: "DeclarationOf", ConstValue: &"100"},
          	"ConstNineNine": &{Name: types.Name{Package: "base/foo/proto", Name: "ConstNineNine"}, Kind: "DeclarationOf", ConstValue: &"99"},
          	... // 4 identical entries
          }
```

I'd say the options are
1. Use `ExactString` always
1. Use `ExactString` if the underlying type is a string, and `String` otherwise
1. Use `String` always

(1) is the most accurate but if anyone actually needs the numeric constants it will be confusing. (3) is out because we definitely don't want strings to be possibly truncated. (2) is Ok for documentation purposes, but maybe not for other uses (not sure what they would be though).

parser/parse_test.go
```diff
@@ -253,6 +274,62 @@ const FooEnumSymbol proto.Enumeration = proto.EnumSymbol
...
+		},
+		"ConstHundred": &types.Type{
+			Name:       types.Name{Package: "base/foo/proto", Name: "ConstHundred"},
+			Kind:       types.DeclarationOf,
+			ConstValue: strPtr("100"),
+		},
+		"ConstFloat": &types.Type{
+			Name:       types.Name{Package: "base/foo/proto", Name: "ConstFloat"},
+			Kind:       types.DeclarationOf,
+			ConstValue: strPtr("8782019273372467/1125899906842624"),
```
lavalamp, line 317: Honestly this seems like a bug in go. That actually works out to 7.79999999999999982236431605997495353221893310546875 and only gets the right answer if you truncate to float64.

https://play.golang.org/p/N6b9whhPoxa

parser/parse_test.go
lavalamp: Let's go with option 2 :/ I knew go represents constants with theoretically infinite precision, but I did not expect it to work like this.

jpeach: OK, I implemented (2) and added some comments. Note that value for string constants is no longer quoted (`ExactString` just calls `strconv.Quote` anyway).

parser/parse_test.go
jpeach: It's pretty close though! 😂

lavalamp:
OK this LGTM but I need you to sign the CLA first--sorry for not noticing previously!

jpeach:
> OK this LGTM but I need you to sign the CLA first--sorry for not noticing previously!

I've signed The Linux Foundation CLA and the CI check is passing. Is that the CLA you meant?

lavalamp:
Yup!

/lgtm
/approve

PR #190, opened 2020-11-02
Preallocate set to the length of the passed elements

markusthoemmes:
When passing a list via the constructor, we usually preallocate "manually" in downstream projects to keep things small'ish.

I do realize that this might be actually worse for lists that are passed for deduplication and where a massive amount of duplicates are present. If we think that's an important case, we could alternative add a new constructor for this behavior or keep doing what we're doing downstream.


lavalamp:
My guess is that preallocating will be a slight improvement, that we don't rely too much on massive deduplication. Anyway, we're generally more worried about number of allocations than total size.

(of course, it's extremely hard to be sure about such things without a benchmark on the consuming function(s) that are affected...)

/lgtm
/approve

PR #192, opened 2020-11-03
Update defaulter-gen to support +default annotation

Jefftree:
Update default-gen to incorporate additions from [defaulting KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1929-built-in-default)

https://github.com/kubernetes/kubernetes/pull/96317 shows an example for adding an annotation to ContainerPort.


examples/defaulter-gen/generators/defaulter.go
```diff
@@ -401,6 +411,18 @@ func newCallTreeForType(existingDefaulters, newDefaulters defaulterFuncMap) *cal
 	}
 }
 
+func populateDefaultValue(node *callNode, commentLines []string) *callNode {
+	// TODO: Check eligibility of defaulting
```
jpbetz, line 415: Options here: 1. replicate what @apelisse did in his OpenAPI PR here, 2. do nothing let the OpenAPI code own the eligibility

WDYT @apelisse ?

apelisse: I'd go with 1., we can't hope that the other system catches the problem, that'd be an implicit dependency between the two.

jpbetz:
@Jefftree Let's add a "annotation" directory to https://github.com/kubernetes/gengo/tree/master/examples/defaulter-gen/_output_tests and include a simple types.go with a simple `+default` example. Looks like after we add the types.go file we can run `make test` (https://github.com/kubernetes/gengo/blob/master/examples/defaulter-gen/Makefile) which will output the the generated code that we can then paste into `zz_generated` generated and check in.

apelisse:
Thanks @jpbetz, I was going to ask how I could see the generated code! That's useful for reviewing.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -731,6 +803,26 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
 	}
 }
 
+func (n *callNode) writeDefaulter(varName string, isVarPointer bool, sw *generator.SnippetWriter) {
+	accessor := varName
+	if !isVarPointer {
+		accessor = "&" + accessor
+	}
+	if n.defaultValue != "" {
+		sw.Do("if reflect.ValueOf($.var$).IsNil() {\n", generator.Args{
```
apelisse, line 812: that's not great, IsNil only works for specific types (pointer and pointer-like types)

apelisse: What about [`IsZero`](https://golang.org/pkg/reflect/#Value.IsZero) instead?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -731,6 +803,26 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
...
+			"var": varName,
+		})
+		sw.Do("if err := json.Unmarshal([]byte(`$.defaultValue$`), $.var$); err != nil {\n", generator.Args{
```
apelisse, line 815: Don't you miss an `&` before `$.var$`?

apelisse: Ah, no ok I can see the accessor above, cool.

apelisse: Maybe use a different name "var" is a little confusing

jpbetz:
@Jefftree I think you already found this, but just so we have it for posterity, here's the code used in OpenAPI to detect the corner case of when a +default annotation has the same value as the zero value of the type: https://github.com/kubernetes/kube-openapi/pull/210/files#diff-40395b83262da9da364bbc5f38969ccd397d3344be594f69203eb62773cab0bdR536

cc @apelisse

examples/defaulter-gen/generators/defaulter.go
Jefftree: Thanks for the catch, updated.

examples/defaulter-gen/generators/defaulter.go
Jefftree: renamed to varPointer

Jefftree:
Updated to address comments and incorporate check for when a +default annotation has the same value as the zero value of the type. 

The code is still having some trouble generating defaults for pointers to structs and I will do some more investigation.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -38,9 +39,54 @@ type CustomArgs struct {
 	ExtraPeerDirs []string // Always consider these as last-ditch possibilities for conversions.
 }
 
+type typeInfo struct {
+	name   string
+	format string
+	zero   interface{}
+}
+
+var schemaTypeFormatMap = map[string]typeInfo{
```
apelisse, line 48: Feel free to drop the type above and make this directly a map of zero-values.
```suggestion
var schemaTypeFormatMap = map[string]interface{}{
```
And then, because that's the only thing in the map, I think it's perfectly fine to also remove the function to access it.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -731,6 +864,26 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
 	}
 }
 
+func (n *callNode) writeDefaulter(varName string, isVarPointer bool, sw *generator.SnippetWriter) {
+	varPointer := varName
+	if !isVarPointer {
+		varPointer = "&" + varPointer
+	}
+	if n.defaultValue != "" {
```
apelisse, line 872: here you need to check if the defaultValue is the zero value for the type too

Jefftree: I don't think a mismatch would be possible here. L501-L504 checks the specific case and breaks out of the function early. The defaultValue field should be empty here if the defaultValue in the comments match the zero value.

apelisse: Awesome!

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -401,6 +452,75 @@ func newCallTreeForType(existingDefaulters, newDefaulters defaulterFuncMap) *cal
...
+			panic(err)
+		} else if enforced != nil {
+			if defaultValue != nil {
+				if reflect.DeepEqual(defaultValue, enforced) {
+					// If the default value annotation matches the default value for the type,
+					// do not generate any defaulting function
+					return node
+				} else {
+					enforcedJSON, _ := json.Marshal(enforced)
+					panic(fmt.Errorf("Invalid default value (%#v) for non-pointer/non-omitempty. If specified, must be: %v", defaultValue, string(enforcedJSON)))
```
apelisse, line 507: This code uses the log package, maybe use `log.Fatal`

examples/defaulter-gen/generators/defaulter.go
Jefftree: Ahh, shouldn't have copied over the function without reading everything. Will update this to just the zero-values map.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	SetObjectDefaults_SubStruct(&in.OtherSub)
+	if reflect.ValueOf(in.Map).IsZero() {
+		if err := json.Unmarshal([]byte(`{"foo": "bar"}`), &in.Map); err != nil {
+			panic(err)
+		}
+	}
+	for i_Map := range in.Map {
+		if reflect.ValueOf(in.Map[i_Map]).IsZero() {
+			i_Map_default := reflect.Zero(in.Map[i_Map])
```
Jefftree, line 78: This statement is necessary because map addresses are not addressable. Thanks Antoine for the suggestion!

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -758,19 +936,31 @@ func (n *callNode) WriteMethod(varName string, depth int, ancestors []*callNode,
...
+		n.writeDefaulter(varName, index, isPointer, sw)
 		n.writeCalls(local, true, sw)
 		for i := range n.children {
 			n.children[i].WriteMethod(local, depth+1, append(ancestors, n), sw)
 		}
 		sw.Do("}\n", nil)
 	case n.key:
+		if n.defaultValue != "" {
+			// Map keys are typed and cannot share the same index variable as arrays and other maps
+			index = index + "_" + ancestors[len(ancestors)-1].field
```
Jefftree, line 956: I think this is the best way to guarantee index variables are unique for map iterations. It's impossible for two variables under the same struct to share the same name

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -473,12 +577,22 @@ func (c *callTreeForType) build(t *types.Type, root bool) *callNode {
...
+			member.index = true
+			member.markerOnly = true
+			parent.children = append(parent.children, *member)
 		}
+
 	case types.Map:
 		if child := c.build(t.Elem, false); child != nil {
 			child.key = true
 			parent.children = append(parent.children, *child)
+		} else if member := populateDefaultValue(nil, t.Elem, "json:\"name,omitempty\"", t.Elem.CommentLines); member != nil {
```
apelisse, line 590: Rather than passing a made-up tag here, I would pass-in the information that you care about in the tags. If I read `populateDefaultValue` properly, you only use the tag to know if the field is omitempty. Can you pass omitempty as a boolean instead of the tag?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -401,6 +436,75 @@ func newCallTreeForType(existingDefaulters, newDefaulters defaulterFuncMap) *cal
...
+			}
+		}
+
+		// callNodes are not automatically generated for primitive types. Generate one if the callNode does not exist
+		if node == nil {
+			node = &callNode{}
+		}
+
+		node.defaultValue = defaultMap[0]
+	} else if len(defaultMap) > 1 {
```
apelisse, line 502: Can you return that error earlier, that'll unnest everything by one-level.

examples/defaulter-gen/generators/defaulter.go
Jefftree: This was meant as a workaround for https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1929-built-in-default#string-maps, but since that case should be invalid we can remove this code path.

examples/defaulter-gen/_output_tests/marker/deepcopy.go
```diff
@@ -0,0 +1,51 @@
+package marker
```
Jefftree, line 1: Unfortunately this file is required in order for the generated code to compile.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -0,0 +1,93 @@
...
+			panic(err)
+		}
+	}
+	for i_Map := range in.Map {
+		if reflect.ValueOf(in.Map[i_Map]).IsZero() {
+			i_Map_default := in.Map[i_Map]
+			if err := json.Unmarshal([]byte(`"apple"`), &i_Map_default); err != nil {
+				panic(err)
+			}
+			in.Map[i_Map] = i_Map_default
```
Jefftree, line 82: I'm not exactly sure why this is required, but the tests fail if this isn't added.

examples/defaulter-gen/_output_tests/marker/defaults.go
```diff
@@ -0,0 +1,31 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func addDefaultingFuncs(scheme *runtime.Scheme) error {
+	return RegisterDefaults(scheme)
+}
+
+func SetDefaults_DefaultedWithFunction(obj *DefaultedWithFunction) {
+	if obj.S1 == "" {
+		obj.S1 = "default_function"
```
apelisse, line 29: the name makes it very obvious 👍

examples/defaulter-gen/tests/marker_test.go
```diff
@@ -142,3 +142,16 @@ func Test_Marker(t *testing.T) {
 		})
 	}
 }
+
+func Test_DefaultingFunction(t *testing.T) {
```
apelisse, line 146: You could have just added an extra field in the other struct, maybe that's simpler? No strong opinion.

Jefftree: Wanted to avoid the defaulting function being executed in the other test since it would apply to every test case and 
 it could be confusing why one of the defaults was changed without any indication in the test name.

apelisse: That makes sense, thanks!

examples/defaulter-gen/Makefile
```diff
@@ -2,7 +2,7 @@ TOOL=defaulter-gen
 
 .PHONY : gen test verify git_verify_clean git_verify_no_change
 test:
-	go test ./...
+	go test ./_output_tests/marker/...
```
apelisse, line 5: why?

Jefftree, line 5: This isn't ideal but none of the other output_tests packages compile without adding the deepcopy and conversion generated code. Given that this PR is pretty big already, I think it's best to do that in a follow up.

Jefftree: This isn't ideal but none of the other output_tests packages compile without adding the deepcopy and conversion generated code. Given that this PR is pretty big already, I think it's best to do that in a follow up.

go.sum
```diff
@@ -1,29 +1,160 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
apelisse, line 1: Do you need to run `go mod tidy`?

examples/defaulter-gen/Makefile
apelisse: Sounds good to me

lavalamp: ugh, I'm so sad this repo grew a makefile :( (not asking for a change)

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -0,0 +1,95 @@
...
+	scheme.AddTypeDefaultingFunc(&Defaulted{}, func(obj interface{}) { SetObjectDefaults_Defaulted(obj.(*Defaulted)) })
+	scheme.AddTypeDefaultingFunc(&DefaultedWithFunction{}, func(obj interface{}) { SetObjectDefaults_DefaultedWithFunction(obj.(*DefaultedWithFunction)) })
+	return nil
+}
+
+func SetObjectDefaults_Defaulted(in *Defaulted) {
+	if reflect.ValueOf(in.Field).IsZero() {
+		in.Field = "bar"
+	}
+	if reflect.ValueOf(in.OtherField).IsZero() {
```
lavalamp, line 43: Hm, these ValueOf calls will probably be slow / make garbage. Can you instead pre-compute the zero value and use == to tell?

Alternatively, you could make a benchmark that says I'm wrong :)

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -0,0 +1,95 @@
...
+		in.OtherField = 0
+	}
+	if reflect.ValueOf(in.List).IsZero() {
+		if err := json.Unmarshal([]byte(`["foo", "bar"]`), &in.List); err != nil {
```
lavalamp, line 47: Hm, we can probably leave this for the first take, but I think we'll eventually want to pre-unmarshal and make shallow or deep copies for these lines as needed.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
apelisse: What about this:
```golang
package test

import (
	"reflect"
	"testing"
)

func Benchmark_Valueof(b *testing.B) {
	b.ReportAllocs()
	for n := 0; n < b.N; n++ {
		if reflect.ValueOf(n + 1).IsZero() {
			b.Fatalf("%v is not 0", n+1)
		}
	}
}

func Benchmark_direct(b *testing.B) {
	b.ReportAllocs()
	for n := 0; n < b.N; n++ {
		if n+1 == 0 {
			b.Fatalf("%v is not 0", n+1)
		}
	}
}
```
Gives:
```shell
$ go test -bench=. ./...
goos: darwin
goarch: amd64
Benchmark_Valueof-8   	72421857	        15.5 ns/op	       8 B/op	       0 allocs/op
Benchmark_direct-8    	1000000000	         0.252 ns/op	       0 B/op	       0 allocs/op
```

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	if reflect.ValueOf(in.OtherSub.I).IsZero() {
+		in.OtherSub.I = 1
+	}
+	if reflect.ValueOf(in.Map).IsZero() {
+		if err := json.Unmarshal([]byte(`{"foo": "bar"}`), &in.Map); err != nil {
+			panic(err)
+		}
+	}
+	for i_Map := range in.Map {
```
lavalamp, line 76: Maybe use `k, v`, then you don't have to do the map lookup.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
lavalamp: Map items are immutable, so you have to do this unless it's a pointer and you can modify the thing it points to.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
lavalamp: hm, if it's really 0 bytes, then maybe this isn't a problem, can you try this with something more complicated like a map, struct, slice... ?

I don't think I care about the CPU time, although that's like 60x slower and these functions call this for every defaulted field (?). 66 fields and we're up to a millisecond, that's a real amount of time.

@wojtek-t @deads2k any thoughts on what our perf budget should be for default setting?

apelisse: In the immediate term, we're only planning on applying this setting to containerPort.protocol, since this is the default that is causing us trouble. As we extend and start using this, we can start optimizing as needed.

> can you try this with something more complicated like a map, struct, slice...

Sure.

lavalamp: OK if we're only using it for one field in 1.20, that should be fine.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
Jefftree: I think we would have to do a map lookup regardless because map values are not addressable (json unmarshal cannot directly be applied) and we would have to create a copy of a value and reassign it back to the map (See L78-82). I'm not sure if this is the best way to do things but it currently fits the cases we are interested in.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
Jefftree: I'll add this as an action item for future improvements.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
Jefftree: Thanks for the benchmarks Antoine! I'll add this as an action item for future improvements/optimizations.

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -0,0 +1,66 @@
...
+	// Default is forced to 0
+	EmptyInt int
+
+	// Default is forced to empty string
+	// Specifying the default is a no-op
+	// +default=""
+	EmptyString string
+
+	// +default=["foo", "bar"]
+	List []Item
```
sttts, line 40: Can I have a `StructList []SubStruct` ?

sttts: And a `PtrStructList []*SubStruct`. Maybe that's less common in real types.

sttts: And a `[]string`.

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -0,0 +1,66 @@
...
+	// Default is forced to empty struct
+	OtherSub SubStruct
+
+	// +default={"foo": "bar"}
+	Map map[string]Item
+}
+
+// +default="apple"
+type Item *string
+
```
sttts, line 53: Is this rejected? Should it?

```
// +default="apple"
type NonPtrItem string
```

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -0,0 +1,66 @@
...
+	List []Item
+	// +default={"s": "foo", "i": 5}
+	Sub *SubStruct
+
+	// Default is forced to empty struct
+	OtherSub SubStruct
+
+	// +default={"foo": "bar"}
+	Map map[string]Item
+}
```
sttts, line 49: what happens with

```
// +default="banana"
Item Item
```

?

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+	if reflect.ValueOf(in.Sub).IsZero() {
+		if err := json.Unmarshal([]byte(`{"s": "foo", "i": 5}`), &in.Sub); err != nil {
+			panic(err)
+		}
+	}
+	if in.Sub != nil {
+		if reflect.ValueOf(in.Sub.I).IsZero() {
+			in.Sub.I = 1
```
sttts, line 65: this surprises me. We don't have a defaulter for SubStruct here and just call it?

sttts: this might to lead to a lot of code for common structs. Not sure how many we have. ObjectMeta comes to mind. Is there much that is defaulted?

apelisse: Yeah, that's a good point

examples/defaulter-gen/_output_tests/marker/type.go
apelisse: This will fail if Item is not a pointer (since you can't have omitempty on the item).

examples/defaulter-gen/_output_tests/marker/type.go
Jefftree: Added test cases for those structures.

examples/defaulter-gen/_output_tests/marker/type.go
Jefftree: By default, I think this should be allowed and it would depend on how other structs use it. For instance (assuming omitempty is not tagged),

```
type S {
   A NonPtrItem
   B *NonPtrItem
}
```

The default "apple" would apply to B because it's a pointer value and A would only be eligible to be defaulted to "".

/cc @apelisse

examples/defaulter-gen/_output_tests/marker/type.go
Jefftree: Added a test case for this. "banana" will be the default value in this case since Item is a pointer.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
apelisse: > Is there much that is defaulted?

 No, not now. As discussed somewhere else, we're only planning on using this for `protocol` right now. We can follow-up on that.

examples/defaulter-gen/_output_tests/marker/type.go
apelisse: Yeah, that's great.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -38,9 +39,38 @@ type CustomArgs struct {
...
+	"byte":        0,
+	"float64":     0.,
+	"float32":     0.,
+	"bool":        false,
+	"time.Time":   "",
+	"string":      "",
+	"integer":     0.,
+	"number":      0.,
+	"boolean":     false,
+	"[]byte":      "", // base64 encoded charactes
```
jpbetz, line 62:
```suggestion
	"[]byte":      "", // base64 encoded characters
```

examples/defaulter-gen/_output_tests/marker/zz_generated.go
apelisse: Can you open an issue and list the things that we need to improve?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -731,6 +899,85 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
 	}
 }
 
+func (n *callNode) writeDefaulter(varName string, index string, isVarPointer bool, sw *generator.SnippetWriter) {
```
apelisse, line 902: Maybe I'm wrong but I suspect this function could be simplified by decomposing the various steps actually involved into separate functions. Let's mark that as a future improvement.

apelisse:
Could you please merge Joe's suggested change and then squash some of these commits? I'm happy to lgtm after that.

Jefftree:
Rebased and squashed some commits. Will also create a tracking issue for the future improvements

examples/defaulter-gen/generators/defaulter.go
Jefftree: I agree. Originally this function was very small but there ended up being quite a number of corner cases to consider.

Jefftree:
Included a tracking issue for future improvements here (https://github.com/kubernetes/gengo/issues/193)

examples/defaulter-gen/_output_tests/marker/zz_generated.go
lavalamp: The test cases don't really cover what happens with types that *aren't* defaulted. Hopefully no functions are emitted / called when no auto defaulting would be done?

lavalamp: I found https://github.com/kubernetes/kubernetes/pull/96317 and that looks awesome, so please add more test cases here for a followup.

examples/defaulter-gen/generators/defaulter.go
jpbetz: +1, I have a couple thoughts as well, but they don't impact correctness, so follow up should be fine

examples/defaulter-gen/_output_tests/marker/zz_generated.go
jpbetz: +1, we've discussed pre-unmarshalling a couple times. It will definitely be on the follow up list.

jpbetz (approved):
/lgtm

This looks sufficient for 1.20 to me, especially since we're only enabling one default field for the release and we know this generates a sane default for it. 

And it looks like we already have a plan to track follow-ups on an issue, the ones I noticed as being most important are:
- organizing/simplifying the code generation
- "pre-unmarshalling" defaults (maybe less urgent given that primitives are generated directly as literals)

thockin:
Hi all.  I know this is "a little old", bu T found my way into this code and am trying to understand.

Can someone explain what the omitempty logic is doing here?  I don't understand what we are trying to express, and why we would look at JSON tags to do that.  It doesn't make sense to me, so I must be missing something.

Jefftree:
> Hi all. I know this is "a little old", but found my way into this code and am trying to understand.
> 
> Can someone explain what the omitempty logic is doing here? I don't understand what we are trying to express, and why we would look at JSON tags to do that. It doesn't make sense to me, so I must be missing something.

https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1929-built-in-default#changes-to-kubebuilderdefault-gen

This KEP probably describes it better than I can

Issue #193, opened 2020-11-12
Tracking Issue for improvements for built-in declarative defaults generator


Jefftree:
Originally PR linked here: https://github.com/kubernetes/gengo/pull/192

Future Improvements:
- Avoid `reflect.ValueOf` and precompute the zero values instead. https://github.com/kubernetes/gengo/pull/192#discussion_r520952124
- Unmarshal once for structs/arrays/maps and use shallow/deep copy instead of unmarshalling every time https://github.com/kubernetes/gengo/pull/192#discussion_r520952702
- Create a defaulting function to call instead of generating the code every time a defaulter is needed https://github.com/kubernetes/gengo/pull/192#discussion_r521637246
- Decompose `writeDefaulter` into separate functions https://github.com/kubernetes/gengo/pull/192#discussion_r522391330
- Remove PruneDefaults from openapi test https://github.com/kubernetes/kubernetes/blob/dee67230af5ee72f616c46856bf5c700d0c9b6b9/pkg/generated/openapi/openapi_test.go#L34

We should target these for the beginning of 1.21

apelisse:
We need to address the comment on [round-tripping the spec through json](https://github.com/kubernetes/kubernetes/blob/master/pkg/generated/openapi/openapi_test.go#L34-L35). This is caused because there is some confusion in the code about what generates integers and what generates floats. Ideally, the type would be explicit everywhere so that they wouldn't be converted by accident.

apelisse:
We need a test to show that we can assign floats properly: https://github.com/kubernetes/gengo/blob/master/examples/defaulter-gen/_output_tests/marker/type.go#L23

Jefftree:
/close

I think we covered all the tasks that we wanted to do.

PR #195, opened 2020-11-19
Precompute zero values and add test for float

Jefftree:
Addresses part of https://github.com/kubernetes/gengo/issues/193

- Precompute zero values and drop using the reflect library
- Add test for defaulting of floats

For floats, checking equality to 0 works the same way as ints (reflect works a bit differently but isn't really applicable here).

```
>>> var i float
>>> var j int
>>> i == 0
True
>>> j == 0
True
```

When checking for zero values, primitive types are checked against their zero value entry in `typeZeroValue` map, while non-primitive types are checked against nil.

A diff of this PR against the k/k types can be found here: https://github.com/kubernetes/kubernetes/pull/96960


examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -24,17 +24,32 @@ type Defaulted struct {
...
-
-	// Default is forced to 0
-	EmptyInt int
+	StringDefault string `json:"omitempty"`
 
 	// Default is forced to empty string
 	// Specifying the default is a no-op
 	// +default=""
-	EmptyString string
+	StringEmptyDefault string
```
apelisse, line 32: Can you add an example of a scalar pointer type?

apelisse:
That generated code looks great!

apelisse:
LGTM once you add the scalar pointer.

examples/defaulter-gen/_output_tests/marker/type.go
Jefftree: Added.

examples/defaulter-gen/_output_tests/marker/zz_generated.go
```diff
@@ -37,79 +36,87 @@ func RegisterDefaults(scheme *runtime.Scheme) error {
...
-		in.Field = "bar"
+	if in.StringDefault == "" {
+		in.StringDefault = "bar"
 	}
-	if reflect.ValueOf(in.OtherField).IsZero() {
-		in.OtherField = 0
+	if in.StringPointer == nil {
+		var ptrVar1 string = "default"
+		ptrVar0 := &ptrVar1
+		in.StringPointer = ptrVar0
```
apelisse, line 45: Can't you remove one indirection here?
```golang
		var ptrVar0 string = "default"
		in.StringPointer = &ptrVar0
```

Jefftree: Fixed.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -899,80 +906,89 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
...
-				}
-				sw.Do("$.varName$[$.index$] = ptrVar0", args)
-			} else {
-				sw.Do("$.varName$[$.index$] = $.defaultValue$", args)
-			}
-		} else {
-			sw.Do("if err := json.Unmarshal([]byte(`$.defaultValue$`), $.varPointer$[$.index$]); err != nil {\n", args)
-			sw.Do("panic(err)\n", nil)
-			sw.Do("}\n", nil)
+	if n.defaultIsPrimitive {
```
apelisse, line 934: it looks like this part can be moved in the same condition below L962? Or even further done, specifically where it's needed.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -899,80 +906,89 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
...
-		if n.defaultIsPrimitive {
-			if n.defaultDepth > 0 {
-				sw.Do("var ptrVar$.varDepth$ $.varType$ = $.defaultValue$\n", args)
-				for i := n.defaultDepth; i > 0; i-- {
-					sw.Do("ptrVar$.ptri$ := &ptrVar$.i$\n", generator.Args{"i": fmt.Sprintf("%d", i), "ptri": fmt.Sprintf("%d", (i - 1))})
-				}
-				sw.Do("$.varName$[$.index$] = ptrVar0", args)
-			} else {
-				sw.Do("$.varName$[$.index$] = $.defaultValue$", args)
+	variablePointer := variablePlaceholder
```
apelisse, line 957: This one is only used in one place below, can you reduce the scope of this variable? L986

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -899,80 +906,89 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
...
-				}
-				sw.Do("$.varName$ = ptrVar0", args)
-			} else {
-				sw.Do("$.varName$ = $.defaultValue$", args)
-			}
+		sw.Do(fmt.Sprintf("if %s == nil {\n", variablePlaceholder), args)
+		// Map values are not directly addressable and we need a temporary variablePlaceholder to do json unmarshalling
+		if n.key {
+			sw.Do("$.mapDefaultVar$ := $.varName$[$.index$]\n", args)
+			sw.Do("if err := json.Unmarshal([]byte(`$.defaultValue$`), &$.mapDefaultVar$); err != nil {\n", args)
```
apelisse, line 984: so if we had a `map[string]*string`, we would be in that specific condition, right? That's not optimal, but I think that's fine for now. Maybe add a comment?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -899,80 +906,89 @@ func (n *callNode) writeCalls(varName string, isVarPointer bool, sw *generator.S
...
-				sw.Do("var ptrVar$.varDepth$ $.varType$ = $.defaultValue$\n", args)
-				for i := n.defaultDepth; i > 0; i-- {
-					sw.Do("ptrVar$.ptri$ := &ptrVar$.i$\n", generator.Args{"i": fmt.Sprintf("%d", i), "ptri": fmt.Sprintf("%d", (i - 1))})
-				}
-				sw.Do("$.varName$ = ptrVar0", args)
-			} else {
-				sw.Do("$.varName$ = $.defaultValue$", args)
-			}
+		sw.Do(fmt.Sprintf("if %s == nil {\n", variablePlaceholder), args)
+		// Map values are not directly addressable and we need a temporary variablePlaceholder to do json unmarshalling
```
apelisse, line 981: I'm curious if this will work if you add a `StructMap map[string]SubStruct` in your test, can you try?

apelisse (commented):
Thanks! I like the direction!

examples/defaulter-gen/generators/defaulter.go
Jefftree: The `*string` is a pointer to a primitive type so it would go down the defaultIsPrimitive code path. This specific condition is reached if we have something like `map[string]SubStruct where the map values are not primitives.

examples/defaulter-gen/generators/defaulter.go
Jefftree: I added tests for both `StructMap map[string]SubStruct` and `PtrStructMap map[string]*SubStruct`.

Jefftree:
/assign @lavalamp 
/assign @smarterclayton 

(since you expressed concerns around using reflect in the original PR)

See https://github.com/kubernetes/kubernetes/pull/96960 for a diff of k/k defaulting code after this is applied.

apelisse:
/lgtm

Thanks!

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -24,17 +24,35 @@ type Defaulted struct {
...
+
+	// Not specifying a default still defaults for non-omitempty
+	StringEmpty string
+
+	// +default="default"
```
lavalamp, line 37: What should these do?

// +default="de\nfault"
// +default="de\"fault"
// +default="de"fault"

Jefftree: `// +default="de\nfault"`: Code generation passes because this is a valid assignment to a string
`// +default="de\"fault"`: Passes because this is a valid assignment to a string
`// +default="de"fault"`: Fails because this breaks our syntax

We attempt an unmarshal in the code before generation. If the unmarshal fails, we return an error and do not generate the code. If the unmarshal is successful, we pretty much assignment the string value verbatim. Eg:

```
    if in.StringDefault == "" {
        in.StringDefault = "de\"fault"
    }
```

apelisse: I think another way to ask this question is: are all json valid strings also valid (and corresponding) go strings? What about "`"?

lavalamp: I guess I was looking for a test that exercised these. I see a call to strconv.Quote but no Unquote anywhere.

Combined with Antoine's point, I think we should say that the default is using JSON strings, and somewhere I should see a place that unmarshalls via JSON and then calls strconv.Quote to get a go-safe form.

Also I can't seem to find the place that would be returning an error to users? Is it preexisting?

lavalamp: Although I'm realizing these questions can be fixed in a follow-up.

lavalamp:
/approve

Please send a follow-up as needed, thanks.

examples/defaulter-gen/_output_tests/marker/type.go
apelisse: > I think we should say that the default is using JSON strings, and somewhere I should see a place that unmarshalls via JSON and then calls strconv.Quote to get a go-safe form.

It's definitely supposed to be json strings in input, and yes normalizing the string somehow is the way to go, thanks!

Jefftree: Originally there were suggestions around normalizing the string to ensure the input is valid. Right now, the code does a one way unmarshal, and if the unmarshal is successful we use the original string without normalization. I can't really think of a test case where a one way unmarshal would pass but a string normalization would result in a different string. I'd be happy to make the change, just want to make sure it's something we need.

PR #204, opened 2021-06-07
Add support for arrays

christarazi:
Signed-off-by: Chris Tarazi <tarazichris@gmail.com>

### DIFF
--- a/namer/namer.go
+++ b/namer/namer.go
@@ -17,7 +17,9 @@ limitations under the License.
 package namer
 
 import (
+	"fmt"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"k8s.io/gengo/types"
@@ -246,6 +248,12 @@ func (ns *NameStrategy) Name(t *types.Type) string {
 			"Slice",
 			ns.removePrefixAndSuffix(ns.Name(t.Elem)),
 		}, ns.Suffix)
+	case types.Array:
+		name = ns.Join(ns.Prefix, []string{
+			"Array",
+			ns.removePrefixAndSuffix(fmt.Sprintf("%d", t.Len)),
+			ns.removePrefixAndSuffix(ns.Name(t.Elem)),
+		}, ns.Suffix)
 	case types.Pointer:
 		name = ns.Join(ns.Prefix, []string{
 			"Pointer",
@@ -340,6 +348,9 @@ func (r *rawNamer) Name(t *types.Type) string {
 		name = "map[" + r.Name(t.Key) + "]" + r.Name(t.Elem)
 	case types.Slice:
 		name = "[]" + r.Name(t.Elem)
+	case types.Array:
+		l := strconv.Itoa(int(t.Len))
+		name = "[" + l + "]" + r.Name(t.Elem)
 	case types.Pointer:
 		name = "*" + r.Name(t.Elem)
 	case types.Struct:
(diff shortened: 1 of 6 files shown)


christarazi:
I tried adding some unit tests to validate this change (see last commit marked as WIP), but running into issues that I don't quite understand:
```
    parse_test.go:719: 4-f.Test: type kind wrong, wanted Array, got Slice (&types.Type{Name:types.Name{Package:"", Name:"[]a.Test", Path:""}, Kind:"Slice", CommentLines:[]string(nil), SecondClosestCommentLines:[]string(nil), Members:[]types.Member(nil), Elem:(*types.Type)(0xc000427680), Key:(*types.Type)(nil), Underlying:(*types.Type)(nil), Methods:map[string]*types.Type(nil), Signature:(*types.Signature)(nil), ConstValue:(*string)(nil), Len:0})
--- FAIL: TestTypeKindParse (0.00s)
FAIL
FAIL    k8s.io/gengo/parser     0.042s
FAIL
```

Is there something I'm missing?

christarazi:
@lavalamp @sttts I would appreciate your input on the above.

types/types.go
```diff
@@ -350,7 +352,8 @@ type Type struct {
 
 	// TODO: Add:
 	// * channel direction
-	// * array length
+
+	Len int64
```
lavalamp, line 356: Add a comment stating what it's used for.

namer/namer.go
```diff
@@ -246,6 +247,11 @@ func (ns *NameStrategy) Name(t *types.Type) string {
 			"Slice",
 			ns.removePrefixAndSuffix(ns.Name(t.Elem)),
 		}, ns.Suffix)
+	case types.Array:
+		name = ns.Join(ns.Prefix, []string{
+			"Array",
```
lavalamp, line 252: This name needs to add the length, perhaps right here.

parser/parse_test.go
```diff
@@ -627,6 +628,10 @@ func TestTypeKindParse(t *testing.T) {
 			Package: "f", Name: "Test", k: types.Slice,
 			names: []string{"Test", "FTest", "test", "fTest", "f.Test"},
 		},
+		{
+			Package: "f", Name: "Test", k: types.Array,
```
lavalamp, line 632: This should be Test5, no?

If this test is passing as written, something is wrong...

namer/namer.go
```diff
@@ -340,6 +346,9 @@ func (r *rawNamer) Name(t *types.Type) string {
 		name = "map[" + r.Name(t.Key) + "]" + r.Name(t.Elem)
 	case types.Slice:
 		name = "[]" + r.Name(t.Elem)
+	case types.Array:
+		l := strconv.Itoa(int(t.Len))
+		name = "[" + l + "]" + r.Name(t.Elem)
```
lavalamp, line 351: Please make sure a test exercises this, I don't see one that does in this change.

namer/namer.go
lavalamp: Also needs to be covered by a test.

lavalamp (commented):
Surprisingly small change, looks mostly good -- tests need some attention.

parser/parse_test.go
christarazi: Moved this under a new "h" pkg

christarazi:
Comments have been addressed, PTAL @lavalamp

namer/namer.go
lavalamp: Changes look good, I have a slight preference for "ArrayOf5Foo" or even just "Array5Foo" because it matches the go ordering of `[5]Foo`. What do you think?

christarazi: "Array5Foo" sounds good to me. Updated.

lavalamp:
/lgtm
/approve

Thanks!

PR #208, opened 2021-08-12
Make defaulter-gen accept package paths as input

liggitt:
this allows using package import paths, needed to work in module mode

xref https://github.com/kubernetes/kubernetes/issues/82531 and https://github.com/kubernetes/kubernetes/pull/104330

/hold for green CI with this change in k/k

* [x] run with existing defaulter-gen tags: https://github.com/kubernetes/kubernetes/pull/104333
* [x] run with defaulter-gen tags switched to import packages: https://github.com/kubernetes/kubernetes/pull/104330

/assign @sttts

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -318,9 +318,18 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 		}
 		if len(inputTags) == 1 {
 			var err error
-			typesPkg, err = context.AddDirectory(filepath.Join(pkg.Path, inputTags[0]))
+
+			inputPath := inputTags[0]
+			if strings.HasPrefix(inputPath, "./") || strings.HasPrefix(inputPath, "../") {
+				// this is a relative dir, which will not work under gomodules.
+				// join with the local package path, but warn
+				klog.Warningf("relative path %s=%s will not work under gomodule mode; use full package path (as used by 'import') instead", inputTagName, inputPath)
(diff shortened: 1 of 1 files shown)


examples/defaulter-gen/generators/defaulter.go
```diff
@@ -318,9 +318,18 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 		}
 		if len(inputTags) == 1 {
 			var err error
-			typesPkg, err = context.AddDirectory(filepath.Join(pkg.Path, inputTags[0]))
+
+			inputPath := inputTags[0]
+			if strings.HasPrefix(inputPath, "./") || strings.HasPrefix(inputPath, "../") {
```
liggitt, line 323: this preserves existing behavior for existing relative-path consumers, but warns they should switch to package mode

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -318,9 +318,18 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
...
+				// this is a relative dir, which will not work under gomodules.
+				// join with the local package path, but warn
+				klog.Warningf("relative path %s=%s will not work under gomodule mode; use import path instead", inputTagName, inputPath)
```
sttts, line 326: as somebody reading this later, what is an "import path"? You want to say "absolute path as used in imports" or "full import path".

sttts:
/lgtm
/approve

/hold

if you want to be more specific in the message. Unhold at will.

liggitt:
updated message

/hold cancel

PR #215, opened 2021-09-27
add module support

liggitt:
- [x] normalizes packages when looking up and storing resolved package info
- [x] add mechanism to resolve multiple packages at once
- [x] optimize for orders-of-magnitude slower performance using module-aware loaders (https://github.com/golang/go/issues/31087#issuecomment-763176827)

Hoisted out of https://github.com/kubernetes/kubernetes/pull/99226

xref https://github.com/kubernetes/kubernetes/issues/82531

Need to ensure this works in k8s.io/kubernetes in GOPATH mode (https://github.com/kubernetes/kubernetes/pull/105299) and module mode (https://github.com/kubernetes/kubernetes/pull/99226) before merge

/cc @sttts
/sig api-machinery


thockin (commented):
I don't get the "add multiple packages at once" commit - it looks like a code cleanup but semantically the same - am I missing something?

Overall I get what you're doing here.  I hope we can rip out a lot of this and just use tools/packages, and drop all the logic we do to recreate things like "pkg/..." support.

thockin:
YOu know what this need - tes cases to make sure we don't backslide

stevekuznetsov:
@liggitt any chance you are still interested in pursuing this, or, if not, may I?

sttts:
@stevekuznetsov

liggitt:
> @liggitt any chance you are still interested in pursuing this, or, if not, may I?

go ahead. some of this was hoisted into https://github.com/kubernetes/gengo/pull/222

also, check with @thockin, who has been looking at alternative ways of making the generators work in module mode

thockin:
I have a k/k branch WIP which makes the codegens work in module mode AND
the new "workspace" mode (go 1.18) with `packges.Load`.  The changes are
"significant" and not ready for integration.  My goal is to make workspaces
work, then I will circle back and clean up the gengo hacks (and they are
indeed hacks).  To hack, I forked it to a local "gengo2" directory.

https://github.com/thockin/kubernetes/tree/go-workspaces/gengo2

Note that the way the Makefile is written tries to run each tool one time
for ALL modules.   This *only* works in workspace mode (which requires go
1.18).  To support modules without workspaces, we would need to invoke each
tool once for each module -- muuuuuuuch slower.

Comments are welcome, but please note the goal was NOT "a clean set of
patches" yet.

in particular:

https://github.com/thockin/kubernetes/commit/05c11254efd5d825ced750958108bbd4b31d70d5
https://github.com/thockin/kubernetes/commit/61bef5c42ccc229ab0967f0e90428c51e14cc6b1
https://github.com/thockin/kubernetes/commit/6c1fc95985916f036c2eb07284b08d17a48886fb

If anything, I hope to land this during the 1.25 cycle, after go 1.18 is
mandated.

PR #218, opened 2021-09-28
Fixes kubernetes#217

jiangyuzhao:
If name has prefix 'func (', just return types.Name without parsing more. Fixes kubernetes#217

### DIFF
--- a/parser/parse_test.go
+++ b/parser/parse_test.go
@@ -35,13 +35,14 @@ import (
 )
 
 func TestRecursive(t *testing.T) {
+	dir := "k8s.io/gengo/testdata/a"
 	d := args.Default()
-	d.InputDirs = []string{"k8s.io/gengo/testdata/a/..."}
+	d.InputDirs = []string{dir + "/..."}
 	b, err := d.NewBuilder()
 	if err != nil {
 		t.Fatalf("Fail making builder: %v", err)
 	}
-	_, err = b.FindTypes()
+	findTypes, err := b.FindTypes()
 	if err != nil {
 		t.Fatalf("Fail finding types: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


jiangyuzhao:
@sttts Cloud you please also take a look?

jiangyuzhao:
@lavalamp @sttts Cloud you pls take a look?

lavalamp:
Also, you'll need to sign the CLA.

jiangyuzhao:
@lavalamp I have signed the CLA and modify the test file. In the old implement it will parse the type error and treat it as package name.

jiangyuzhao:
@lavalamp  Cloud you pls take a look?

jiangyuzhao:
@lavalamp hello?

parser/parse_test.go
```diff
@@ -62,6 +62,10 @@ func TestRecursive(t *testing.T) {
 	if foundC {
 		t.Error("Did not expect to find package c")
 	}
+	// only has three package: package "a", package "b", and package "" for all
+	if len(findTypes) != 3 {
```
lavalamp, line 66: I was expecting a check that the type identified is actually correct? From my reading the wrong code was still making a type, it's just not the right construction.

lavalamp:
BTW we are in extremely different time zones, and I do this during working hours, so you can't expect a fast response, sorry

jiangyuzhao:
@lavalamp Sorry for disturbing you at the weekend, I just thought it's daytime and forgot it's also weekend.  
Now I update the commit for type check, but it's somehow ugly code. I just found that if I check the length of `findTypes` which's keys are package names is ok, in the wrong code it will get keys like `func (*k8s.io/gengo/testdata/a.AA).AFunc(i *int, j int) (*k8s.io/gengo/testdata/a.A, k8s.io/gengo/te` which I don't know how it's parsed.

lavalamp:
OK, it seems possible that this could break (in a different way) and the test could still pass, but this is probably good enough for now.

/lgtm
/approve

PR #222, opened 2021-11-22
normalize loaded build packages and handle missing packages without panicking

liggitt:
/hold for clean run in k/k

### DIFF
--- a/parser/parse.go
+++ b/parser/parse.go
@@ -52,7 +52,9 @@ type Builder struct {
 	// This might hold the same value for multiple names, e.g. if someone
 	// referenced ./pkg/name or in the case of vendoring, which canonicalizes
 	// differently that what humans would type.
-	buildPackages map[string]*build.Package
+	//
+	// This must only be accessed via getLoadedBuildPackage and setLoadedBuildPackage
+	buildPackages map[importPathString]*build.Package
 
 	fset *token.FileSet
 	// map of package path to list of parsed files
@@ -102,7 +104,7 @@ func New() *Builder {
 	c.CgoEnabled = false
 	return &Builder{
 		context:               &c,
-		buildPackages:         map[string]*build.Package{},
+		buildPackages:         map[importPathString]*build.Package{},
 		typeCheckedPackages:   map[importPathString]*tc.Package{},
 		fset:                  token.NewFileSet(),
 		parsed:                map[importPathString][]parsedFile{},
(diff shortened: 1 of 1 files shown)


parser/parse.go
```diff
@@ -118,11 +120,37 @@ func (b *Builder) AddBuildTags(tags ...string) {
 	b.context.BuildTags = append(b.context.BuildTags, tags...)
 }
 
+func (b *Builder) getLoadedBuildPackage(importPath string) (*build.Package, bool) {
+	canonicalized := canonicalizeImportPath(importPath)
+	if string(canonicalized) != string(importPath) {
```
thockin, line 125: why are you casting `importPath` to string ?  It's already a string...

parser/parse.go
```diff
@@ -118,11 +120,37 @@ func (b *Builder) AddBuildTags(tags ...string) {
...
+		klog.V(5).Infof("getLoadedBuildPackage: %s normalized to %s", importPath, canonicalized)
+	}
+	buildPkg, ok := b.buildPackages[canonicalized]
+	return buildPkg, ok
+}
+func (b *Builder) setLoadedBuildPackage(importPath string, buildPkg *build.Package) {
+	canonicalizedImportPath := canonicalizeImportPath(importPath)
+	if string(canonicalizedImportPath) != string(importPath) {
```
thockin, line 133: Same - why cast?

parser/parse.go
```diff
@@ -118,11 +120,37 @@ func (b *Builder) AddBuildTags(tags ...string) {
...
+		klog.V(5).Infof("setLoadedBuildPackage: importPath %s normalized to %s", importPath, canonicalizedImportPath)
+	}
+
+	canonicalizedBuildPkgImportPath := canonicalizeImportPath(buildPkg.ImportPath)
+	if string(canonicalizedBuildPkgImportPath) != string(buildPkg.ImportPath) {
```
thockin, line 138: This cast also feels suspect - are they both `importPathString`?  Why not just compare them?

parser/parse.go
```diff
@@ -118,11 +120,37 @@ func (b *Builder) AddBuildTags(tags ...string) {
...
+		klog.V(5).Infof("setLoadedBuildPackage: buildPkg.ImportPath %s normalized to %s", buildPkg.ImportPath, canonicalizedBuildPkgImportPath)
+	}
+
+	if string(canonicalizedImportPath) != string(canonicalizedBuildPkgImportPath) {
```
thockin, line 142: same

thockin (commented):
I think I grok what this is doing.  We have a distinct lack of testing here, don't we...

parser/parse.go
liggitt: `buildPkg.ImportPath` is a string (it's not our type), but I dropped the unnecessary cast

thockin (commented):
Thanks!

/lgtm
/approve

parser/parse.go
```diff
@@ -287,8 +313,8 @@ func (b *Builder) AddDirectoryTo(dir string, u *types.Universe) (*types.Package,
 	if _, err := b.importPackage(dir, true); err != nil {
 		return nil, err
 	}
-	pkg, ok := b.buildPackages[dir]
-	if !ok {
+	pkg, ok := b.getLoadedBuildPackage(dir)
+	if !ok || pkg == nil {
```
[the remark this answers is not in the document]
liggitt: there were a bunch of places where we were looking up directly in the map and assuming non-nil... this was just one

PR #223, opened 2022-01-19
Read package comments from groupversion_info.go for kubebuilder

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>

Kubebuilder generated api types will have the groupname annotation in the `groupversion_info.go` file. This change ensures that it is possible to run gengo against kubebuilder generated codebase to correctly detect groupName.


apelisse:
cc @Jefftree @kevindelgado maybe?

kevindelgado:
I think this makes sense, /lgtm 
can you show a quick before and after screenshot of where this change shows up?

tamalsaha:
Sorry, I don't have a screenshot  But I was trying to use https://github.com/ahmetb/gen-crd-api-reference-docs against CRDs generated using kubebuilder. It did not work with this pr.

PR #227, opened 2022-06-08
Add .idea/ to .gitignore

SimonBaeumer:
Signed-off-by: Simon Baeumer <sbaumer@redhat.com>


.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
lavalamp, line 1: Please add a comment to explain what this is (see format of https://github.com/kubernetes/kubernetes/blob/master/.gitignore).

SimonBaeumer: Copied over all IDE related ignore

SimonBaeumer: Thanks for the review.
Copied over all IDE related ignore config from the kubernetes repo and OSX files.

PR #228, opened 2022-07-01
Update go.mod to match pinned versions in k/k

justinsb:
Currently github.com/kubernetes/code-generator (for example) does not
build due to version skew problems due to the old version of
golang/x/tools


justinsb:
The current pins are incompatible with go workspaces, anyway; I _guess_ that replace is considered with go workspaces but ignored with go imports (?)

PR #230, opened 2022-07-01
WIP: Add github workflow for CI

justinsb:
The equivalent of the travis workflow.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,28 @@
+---
+name: CI
+
+'on':
+  - push
+  - pull_request
+
+env:
+  GOPROXY: https://proxy.golang.org
+
+jobs:
+  verify:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18.1
+
+      - uses: actions/checkout@v3
+
+      - name: go test
+        run: |
+          go test -v ./...
+
+      - name: verify examples
+        run: |
+          go run ./examples/import-boss/main.go -i $(go list k8s.io/gengo/... | grep -v import-boss/tests | paste -sd',' -) --verify-only


aojea:
it seems that there is no CI gating in this repo, is there @lavalamp ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  GOPROXY: https://proxy.golang.org
+
+jobs:
+  verify:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Set up go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18.1
```
aojea, line 18:
```suggestion
    strategy:
      fail-fast: false
      matrix:
        go-version: [1.18.x, 1.19.x]
    runs-on: ubuntu-20.04
    steps:
      - name: Set up go
        uses: actions/setup-go@v3
        with:
          go-version: ${{ matrix.go-version }}
```

aojea:
/assign @thockin @lavalamp 

I think this repo is merging without running the tests first :thinking:

lavalamp:
heh this explains some things, sorry I didn't see this when it was sent.

/approve

Comment looks worth addressing? (I don't know how this file works)

thockin:
Are we using GH actions for "small" repos or would it be better to teach Prow?

aojea:
My experience for small repos is that actions is faster and simpler

aojea:
> Are we using GH actions for "small" repos or would it be better to teach Prow?

we can do both too https://github.com/kubernetes/test-infra/pull/27586

aojea:
/unassign @wojtek-t @ameukam @thockin @lavalamp

ups, wrong tab sorry

aojea:
/close

Prow config has merged and is working, I think there is no need to duplicate efforts

https://github.com/kubernetes/test-infra/pull/27586

PR #231, opened 2022-08-15
add method SymmetricDifference in set gen

weilaaa:
add method SymmetricDifference in set gen


examples/set-gen/generators/sets.go
```diff
@@ -255,22 +255,32 @@ func (s $.type|public$) Clone() $.type|public$ {
...
 	}
 	return result
 }
 
+// SymmetricDifference returns a set of elements which are in either of the sets, but not in their intersection.
+// For example:
+// s1 = {a1, a2, a3}
+// s2 = {a1, a2, a4, a5}
+// s1.Difference(s2) = {a3, a4, a5}
+// s2.Difference(s1) = {a3, a4, a5}
```
MikeSpreitzer, line 279: typo here, I think you want SymmetricDifference

MikeSpreitzer (commented):
Looks good except for a typo noted inline

MikeSpreitzer (approved):
/LGTM

weilaaa:
@lavalamp PTAL

lavalamp:
Looks good but I think you need to actually run this and check in the results?

lavalamp:
(eventually this will all be replaced by type parameters, btw)

weilaaa:
> Looks good but I think you need to actually run this and check in the results?

I made a unit test in another [111768](https://github.com/kubernetes/kubernetes/pull/111768/files) to check it out.

weilaaa:
> (eventually this will all be replaced by type parameters, btw)
Yes, I really think so as I point it in issues [#111853](https://github.com/kubernetes/kubernetes/issues/111853)

lavalamp:
> I made a unit test in another pr [#111768](https://github.com/kubernetes/kubernetes/pull/111768/files) to check it out.

It should go in the same PR, and the other set types should be updated, not just string. I thought we had a test that would ensure changes like this wouldn't be checked in with skew.

weilaaa:
> > I made a unit test in another pr [#111768](https://github.com/kubernetes/kubernetes/pull/111768/files) to check it out.
> 
> It should go in the same PR, and the other set types should be updated, not just string. I thought we had a test that would ensure changes like this wouldn't be checked in with skew.

@lavalamp I generated the set code and made a unit test for it. PTAL

examples/set-gen/sets/set_test.go
```diff
@@ -118,6 +118,25 @@ func TestStringSetDifference(t *testing.T) {
...
 
+func TestStringSetSymmetricDifference(t *testing.T) {
+	a := NewString("1", "2", "3")
+	b := NewString("1", "2", "4", "5")
+	c := a.SymmetricDifference(b)
+	d := b.SymmetricDifference(a)
+	if len(c) != 3 {
+		t.Errorf("Expected len=3: %d", len(c))
+	}
+	if !c.Equal(NewString("3", "4", "5")) {
```
lavalamp, line 129: Nit: equal already checks the length for you, you don't need to check it separately.

lavalamp:
looks good, minor comment on the test.

weilaaa:
> looks good, minor comment on the test.

Thanks for point it, fixed.

PR #232, opened 2022-08-27
allow +default to refer to symbols from code

alexzielenski:
implements the idea from here: https://github.com/kubernetes/kubernetes/issues/104207

basically adds support for following marker syntax:

Local Package Reference:
```
// +default=ref(AConstantFromThisPackage)
```

Fully Qualified Name:
```
// +default=ref(k8s.io/api/core/v1.TerminationMessagePathDefault)
```

Since the new `ref(` prefix is not valid JSON (a requirement of the original syntax), this change does not change behavior of any previously valid marker comments.

cc @thockin @apelisse 

When a Fully Qualified Symbol name is added, adds necessary imports

Caveats:
1. The symbol is not checked for existence at generation-time. Generation is entirely string manipulation-based
2. In the same vein, the type is not checked at generation-time


examples/defaulter-gen/generators/defaulter.go
```diff
@@ -497,6 +498,24 @@ func mustEnforceDefault(t *types.Type, depth int, omitEmpty bool) (interface{},
 	}
 }
 
+var refRE = regexp.MustCompile(`^ref\((("(?P<pkg>.*)")\.)?(?P<ident>[^"]*)\)$`)
+var refREPkgIndex = refRE.SubexpIndex("pkg")
+var refREIdentIndex = refRE.SubexpIndex("ident")
+
+// parseAsRef looks for strings that match one of the following:
+//   - ref(Ident)
+//   - ref("pkgpath".Ident)
```
apelisse, line 507: the doc and regex use `DQ "` around the package path but the example code above doesn't? Which is it?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -497,6 +498,24 @@ func mustEnforceDefault(t *types.Type, depth int, omitEmpty bool) (interface{},
 	}
 }
 
+var refRE = regexp.MustCompile(`^ref\((("(?P<pkg>.*)")\.)?(?P<ident>[^"]*)\)$`)
```
apelisse, line 501: I think you have an extraneous capture group.
What about this:
```suggestion
var refRE = regexp.MustCompile(`^ref\(((?P<pkg>.*)\.)?(?P<ident>[^"]*)\)$`)
```

namer/namer.go
```diff
@@ -300,6 +300,7 @@ func (ns *NameStrategy) Name(t *types.Type) string {
 // import. You can implement yourself or use the one in the generation package.
 type ImportTracker interface {
 	AddType(*types.Type)
+	AddSymbol(types.Name)
```
apelisse, line 303: is this going to break existing users of this code? Do we want to create a separate interface?

examples/defaulter-gen/generators/defaulter.go
apelisse: or something simpler like this since we don't really need the split:
```
var refRE = regexp.MustCompile(`^ref\((?P<reference>[^"])\)$`)
```

alexzielenski:
comments addressed

thockin:
nice!

thockin:
I'm not super familiar with those code any more, so I will Antoine own the apporoval, but thanks for doing this!

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -497,6 +498,23 @@ func mustEnforceDefault(t *types.Type, depth int, omitEmpty bool) (interface{},
 	}
 }
 
+var refRE = regexp.MustCompile(`^ref\((?P<reference>[^"]+)\)$`)
+var refREIdentIndex = refRE.SubexpIndex("reference")
+
+// parseAsRef looks for strings that match one of the following:
+//   - ref(Ident)
+//   - ref("pkgpath".Ident)
```
apelisse, line 506: comment is now out of date?
```suggestion
//   - ref(pkgpath.Ident)
```

apelisse (commented):
LGTM after the small comment nit.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -862,6 +908,28 @@ type callNode struct {
 	defaultType string
 }
 
+type defaultValue struct {
+	// The value was written directlyin the marker comment and
```
lavalamp, line 912:
```suggestion
	// The value was written directly in the marker comment and
```

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -862,6 +908,28 @@ type callNode struct {
...
+	// has been parsed as JSON
+	InlineConstant string
+	// The name of the symbol relative to the parsed package path
+	// i.e. k8s.io/pkg.apis.v1.Foo if from another package of simply `Foo`
```
lavalamp, line 916:
```suggestion
	// i.e. k8s.io/pkg.apis.v1.Foo if from another package or simply `Foo`
```

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -766,6 +794,24 @@ func (g *genDefaulter) GenerateType(c *generator.Context, t *types.Type, w io.Wr
 	}
 	i := 0
 	callTree.VisitInOrder(func(ancestors []*callNode, current *callNode) {
+		if len(current.defaultValue.SymbolReference) > 0 {
+			// If the defaultValue was a reference to a symbol instead of a constant,
+			// make sure to add it to imports and resolve and name of the symbol
```
lavalamp, line 799:
```suggestion
			// make sure to add it to imports and resolve the name of the symbol
```

?

lavalamp:
/approve
/hold

(in case you want to fix the typos)

apelisse:
Thanks
/lgtm

thockin:
I started trying to use this and hit some problems.  Will file new issues.

thockin:
It also looks like openapi-gen reads these default tags and explodes - was that updated to match?

```
+++ [1007 09:52:13] Building go targets for linux/amd64
    k8s.io/kube-openapi/cmd/openapi-gen (non-static)
+++ [1007 09:52:19] Generating openapi code for KUBE
E1007 09:52:25.040033  529715 openapi.go:321] Error when generating: defaultMode, DefaultMode *int32
2022/10/07 09:52:25 OpenAPI code generation error: Failed executing generator: some packages had errors:
failed to generate default in k8s.io/api/core/v1.ConfigMapVolumeSource: DefaultMode: failed to unmarshal default: invalid character 'r' looking for beginning of value
!!! [1007 09:52:25] Call tree:
!!! [1007 09:52:25]  1: ./hack/update-codegen.sh:832 codegen::openapi(...)
```

apelisse:
I was looking for the code in `kube-openapi`, but this is not kube-openapi, we need something there too.

apelisse:
That's not going to happen this release unfortunately :-/ Let's revisit after code-freeze.

PR #199, opened 2021-04-23
Adjust log level for uncopyable types

rainest:
Adjust the log level used when deepcopy-gen encounters a type that requests code generation, but does not support it. Both logs for this condition are now Info V1.

deepcopy-gen performs a check for types that request deepcopy-gen but do not support it in several locations:
- When [checking to see if a package requires code generation at all](https://github.com/kubernetes/gengo/blob/de9496dff47bef4ff8f45a56096bf50798cba2a5/examples/deepcopy-gen/generators/deepcopy.go#L177-L186). This currently logs a fatal error if it encounters an invalid request.
- When [checking to see which individual types in a package need code generation](https://github.com/kubernetes/gengo/blob/de9496dff47bef4ff8f45a56096bf50798cba2a5/examples/deepcopy-gen/generators/deepcopy.go#L270-L273). This currently logs at Info V2 and skips over the type (it treats the type as if it was not marked for deepcopy generation): `I0423 10:08:10.379180 3412441 deepcopy.go:271] Type package/foo.bar is not copyable`

The package-level check effectively chooses a single type from the package at random: the range order is not guaranteed, and it may find a type that requests generation and _is_ copyable first, in which case it will proceed generating code for the package. If there are many copyable types and a few uncopyable types, deepcopy-gen will usually succeed (but not generate any code for the uncopyable types) but sometimes fail, which is a rather confusing UX.

Logging at V1 seems appropriate here seems appropriate. Users should be aware if they've made an impossible request (so that they can either make the type copyable or explicitly disable code generation for it). Since the invalid configuration is handled gracefully (those types are skipped), the package-level check should continue searching for other copyable types in the package rather than failing abruptly.

PR #233, opened 2022-09-21
Restore uncopyable type case in Filter

rainest:
Restore the false return for uncopyable types in generators.Filter(). This is necessary to avoid feeding non-struct types into the generator. This partially reverts [a change](https://github.com/kubernetes/gengo/pull/199/files#diff-b7ebf0c01247d2b5d81765758541f8bc7f03860ae4727278f49ee89444e5af81L270-L273) from #199 based on @thockin's [reported issue](https://github.com/kubernetes/gengo/pull/199#issuecomment-1251674636).

I don't quite understand the other types that otherwise get sent to the generator without this, but generally it looks like they're all [non-structs](https://gist.github.com/rainest/2d53d7c6924f859a7a40e8053156377a#file-uncopyable-txt). These aren't included in the initial package scan but do attempt deepcopy generation after AFAICT.

This change allows generation to proceed on k/k while [still aggregating all the uncopyable struct types](https://gist.github.com/rainest/2d53d7c6924f859a7a40e8053156377a#file-test-txt) during the initial package scan (it lists all problem types in [my code that prompted the original change](https://github.com/kubernetes/gengo/pull/199#issuecomment-1048319739)).

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -273,6 +273,10 @@ func (g *genDeepCopy) Filter(c *generator.Context, t *types.Type) bool {
 	if !enabled {
 		return false
 	}
+	if !copyableType(t) {
+		klog.V(2).Infof("Type %v is not copyable", t)
+		return false
+	}
 	klog.V(4).Infof("Type %v is copyable", t)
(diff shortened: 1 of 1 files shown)


thockin:
Thanks!

/lgtm
/approve

thockin:
I guess I can't approve in here any more :)

PR #236, opened 2022-12-27
Move method near the only user of it, fix impl

thockin:
`arguments.InputIncludes()` is only used by import-boss, so this moves it into that package.  This is part of a larger cleanup to enable go workspaces.

In moving it, I beefed up the test and found the impl was a little loose, so I fixed that, too.

k/k import-boss still runs to success with this patch.


thockin:
ping on this.

I am trying to peel out small, self-contained PRs to make go-workspaces work simpler.  This is one.

lavalamp:
Yeah sorry for delay, I wasn't reading my email on Dec 27th, must have missed this when I came back

/lgtm
/approve

thockin:
I may have a few more, I'll propose a re-vendor later :)

Issue #235, opened 2022-10-07
generated defaults lose type information for aliases


thockin:
xref #232 

We have many enum-string fields in our APIs, for example:

```
type ServiceInternalTrafficPolicyType string

const (
    ServiceInternalTrafficPolicyCluster ServiceInternalTrafficPolicyType = "Cluster"
    ServiceInternalTrafficPolicyLocal ServiceInternalTrafficPolicyType = "Local"
)
```

I tried to use this as:

```
    // +default=ref(ServiceInternalTrafficPolicyCluster)
    InternalTrafficPolicy *ServiceInternalTrafficPolicyType 
```

The code that got generated looks like:

```
    if in.Spec.InternalTrafficPolicy == nil {
        var ptrVar1 string = ServiceInternalTrafficPolicyCluster
        in.Spec.InternalTrafficPolicy = &ptrVar1                                                                                                     
    }    
```

The problem is that `ServiceInternalTrafficPolicyCluster` is not type `string`, so it fails to compile.

I see in gengo/examples/defaulter-gen/generators/defaulter.go `callTreeForType.build()` it takes types of kind `Alias` and unpacks them, recursively, but that just can't work - we have to carry the alias type name into the generated code.

The fix is not obvious to me and I am out of time to track it for today.

nikhita:
/assign

created #239

thockin:
This one is still open.  The simple fix:

```
diff --git a/vendor/k8s.io/gengo/examples/defaulter-gen/generators/defaulter.go b/vendor/k8s.io/gengo/examples/defaulter-gen/generators/defaulter.go
index 48566aee542..ff0b634a0e3 100644
--- a/vendor/k8s.io/gengo/examples/defaulter-gen/generators/defaulter.go
+++ b/vendor/k8s.io/gengo/examples/defaulter-gen/generators/defaulter.go
@@ -1031,7 +1033,7 @@ func (n *callNode) writeDefaulter(varName string, index string, isVarPointer boo
                // keep using the address-of operator on the primitive value until the types match
                if n.defaultDepth > 0 {
                        sw.Do(fmt.Sprintf("if %s == nil {\n", variablePlaceholder), args)
-                       sw.Do("var ptrVar$.varDepth$ $.varType$ = $.defaultValue$\n", args)
+                       sw.Do("ptrVar$.varDepth$ := $.defaultValue$\n", args)
                        // We iterate until a depth of 1 instead of 0 because the following line
[...]

PR #239, opened 2023-01-18
Fix declarative defaults for aliases

nikhita:
Fixes https://github.com/kubernetes/gengo/issues/235
For https://github.com/kubernetes/kubernetes/issues/104207

This commit ensures that the alias types are not unpacked to the underlying type in defaulter-gen. Instead the alias type name is carried into the generated code.

For the example described in #235:

Adding the declarative default as:

```go
// +default=ref(k8s.io/api/core/v1.ServiceInternalTrafficPolicyCluster)
InternalTrafficPolicy *ServiceInternalTrafficPolicy `json:"internalTrafficPolicy,omitempty" protobuf:"bytes,22,opt,name=internalTrafficPolicy"`
```

would generate code like:

```go
if in.Spec.InternalTrafficPolicy == nil {
  var ptrVar1 v1.ServiceInternalTrafficPolicy = v1.ServiceInternalTrafficPolicyCluster
  in.Spec.InternalTrafficPolicy = &ptrVar1
}
```

/cc @apelisse @thockin 
/assign @apelisse

### DIFF
--- a/examples/defaulter-gen/generators/defaulter.go
+++ b/examples/defaulter-gen/generators/defaulter.go
@@ -447,9 +447,7 @@ func resolveTypeAndDepth(t *types.Type) (*types.Type, int) {
 	depth := 0
 	for prev != t {
 		prev = t
-		if t.Kind == types.Alias {
-			t = t.Underlying
-		} else if t.Kind == types.Pointer {
+		if t.Kind == types.Pointer {
 			t = t.Elem
 			depth += 1
 		}
@@ -466,9 +464,7 @@ func getNestedDefault(t *types.Type) string {
 		if len(defaultMap) == 1 && defaultMap[0] != "" {
 			return defaultMap[0]
 		}
-		if t.Kind == types.Alias {
-			t = t.Underlying
-		} else if t.Kind == types.Pointer {
+		if t.Kind == types.Pointer {
 			t = t.Elem
 		}
 	}
@@ -480,7 +476,7 @@ func mustEnforceDefault(t *types.Type, depth int, omitEmpty bool) (interface{}, 
 		return nil, nil
 	}
 	switch t.Kind {
-	case types.Pointer, types.Map, types.Slice, types.Array, types.Interface:
+	case types.Pointer, types.Map, types.Slice, types.Array, types.Interface, types.Alias:
 		return nil, nil
 	case types.Struct:
 		return map[string]interface{}{}, nil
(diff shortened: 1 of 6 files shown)


examples/defaulter-gen/generators/defaulter.go
```diff
@@ -562,8 +558,20 @@ func populateDefaultValue(node *callNode, t *types.Type, tags string, commentLin
 		node.markerOnly = true
 	}
 
-	node.defaultIsPrimitive = t.IsPrimitive()
```
MadhavJivrajani, line 565: Unrelated diff? Or is `node` or `t` being effected by the added codepath?

MadhavJivrajani (commented):
This looks great, thank you!

examples/defaulter-gen/generators/defaulter.go
nikhita: This line was moved to line 574. The diff just renders as this line being removed :/

nikhita:
/hold
tests aren't running in CI, created https://github.com/kubernetes/gengo/pull/240 to fix it

apelisse:
cc @alexzielenski Want to take a look?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -466,9 +464,7 @@ func getNestedDefault(t *types.Type) string {
 		if len(defaultMap) == 1 && defaultMap[0] != "" {
 			return defaultMap[0]
 		}
-		if t.Kind == types.Alias {
-			t = t.Underlying
-		} else if t.Kind == types.Pointer {
+		if t.Kind == types.Pointer {
```
alexzielenski, line 467: Changing this logic seems to remove the ability for a default attached to an alias's underlying to be used. Is that correct?

alexzielenski (changes requested):
had a few questions about LocalNameOf, as well as some other small details

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -562,8 +558,20 @@ func populateDefaultValue(node *callNode, t *types.Type, tags string, commentLin
...
+	if t.Kind == types.Alias {
+		parsedName := types.ParseFullyQualifiedName(t.Name.String())
+		tracker := namer.NewDefaultImportTracker(types.Name{})
+		localPackage := generator.GolangTrackerLocalName(&tracker, t.Name)
+		if len(localPackage) > 0 {
+			node.defaultType = localPackage + "." + parsedName.Name
+		} else {
+			node.defaultType = parsedName.Name
+		}
+		node.underlying = t.Underlying
```
alexzielenski, line 571: Does this work when there is an alias of an alias? I think its worth adding a test for that case and a depth of 3 for good measure.

examples/defaulter-gen/generators/defaulter.go
thockin: I want to argue that this FEELS correct (the same way that `type Y X` - Y doesn't inherit X methods, it shouldn't inherit defaults), but I don't know if that would break anything in k/k without actually trying.

thockin:
cat_begging

apelisse:
/close
Replaced by #246

PR #240, opened 2023-01-18
examples/defaulter-gen: be able to run tests via go test

nikhita:
Files whose names begin with "_" are ignored by go test so none of the defaulter-gen tests were running in CI.

This commit renames `_output_tests` to `output_tests` so that tests are run.

After doing so, tests failed because some packages missed a `fake_deepcopy_conversion.go` file. This is necessary because types registered in the scheme need to implement the `DeepCopyObject()` method.

Additionally, some of the tests import `k8s.io/api` and `k8s.io/apimachinery` so these are also added to `go.mod`.


go.mod
```diff
@@ -1,17 +1,32 @@
...
 	github.com/google/gofuzz v1.1.0
-	github.com/kr/pretty v0.2.0 // indirect
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/tools v0.0.0-20200505023115-26f46d2f7ef8
-	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
-	k8s.io/klog/v2 v2.2.0
-	sigs.k8s.io/yaml v1.2.0
+	golang.org/x/tools v0.2.0
+	k8s.io/api v0.26.0
+	k8s.io/apimachinery v0.26.0
```
nikhita, line 12: gengo depending on api and apimachinery doesn't sound good...

cc @liggitt @dims

liggitt:
> Files whose names begin with "_" are ignored by go test so none of the defaulter-gen tests were running in CI.

IIRC, the ignoring was intentional so we didn't create a dependency cycle

examples/defaulter-gen/Makefile
```diff
@@ -2,11 +2,11 @@ TOOL=defaulter-gen
 
 .PHONY : gen test verify git_verify_clean git_verify_no_change
 test:
-	go test ./_output_tests/marker/...
+	go test ./output_tests/marker/...
```
liggitt, line 5: would a different formulation work, like `cd _output_tests/marker && go test ./...`?

go.mod
dims: @nikhita is it possible to introduce a go.mod just for the example?

PR #244, opened 2023-07-26
Normalize generator name in comment between Windows and Linux/Unix

### LINKED ISSUE #242: should use the filepath package, otherwise it won't work on Windows.
https://github.com/kubernetes/gengo/blob/ab3349d207d4cc9f4bc5c37ba6aa062b4b2ffacc/args/args.go#LL124C20-L124C24

danegsta:
The generatorName value was using path.Base instead of filepath.Base, resulting in the full path to the generator being added to the `// Code generated by ...` comment on Windows. This PR switches that to use filepath.Base. Additionally, the generators on Windows have .exe extensions, which would otherwise get included in the comment. To normalize the output between Windows and Linux/Unix, I'm stripping the extension (if present) from the generator name.

The main goal of all this is to ensure that a generator run on Windows doesn't produce different output than running the same generator on Linux/Unix.

This should resolve #242.

### DIFF
--- a/args/args.go
+++ b/args/args.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


thockin:
Thanks!

/lgtm
/approve

PR #246, opened 2023-08-17
Fix declarative defaults using aliases

alexzielenski:
#239 seems to have gone stale so I took a stab at fixing this myself.

One distinction this approach taken here has over the original attempt is that it is able to construct arbitrary interleavings of aliases and pointers from any default convertible to the same base type. 

Based on https://github.com/kubernetes/gengo/pull/247 since it fixes bug resolving local package names


examples/defaulter-gen/generators/defaulter.go
```diff
@@ -520,15 +520,16 @@ func populateDefaultValue(node *callNode, t *types.Type, tags string, commentLin
 	var defaultString string
 	if len(defaultMap) == 1 {
 		defaultString = defaultMap[0]
+	} else if len(defaultMap) > 1 {
```
alexzielenski, line 523: Moved this line up from below since `defaultMap` is not ever mutated

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -520,15 +520,16 @@ func populateDefaultValue(node *callNode, t *types.Type, tags string, commentLin
...
-	if depth > 0 && defaultString == "" {
+	baseT, depth := resolveTypeAndDepth(t)
+	if len(depth) > 0 && defaultString == "" {
 		defaultString = getNestedDefault(t)
 	}
-	if len(defaultMap) > 1 {
-		klog.Fatalf("Found more than one default tag for %v", t.Kind)
-	} else if len(defaultMap) == 0 {
+
+	if len(defaultString) == 0 {
```
alexzielenski, line 532: Changed `len(defaultMap)` -> len(defaultString)` since that seemed to be the original intention to allow nested default to work. Nested default is only used if default map is empty, so the original condition here would block their use.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -520,15 +520,16 @@ func populateDefaultValue(node *callNode, t *types.Type, tags string, commentLin
 	var defaultString string
 	if len(defaultMap) == 1 {
 		defaultString = defaultMap[0]
+	} else if len(defaultMap) > 1 {
+		klog.Fatalf("Found more than one default tag for %v", t.Kind)
 	}
 
-	t, depth := resolveTypeAndDepth(t)
-	if depth > 0 && defaultString == "" {
+	baseT, depth := resolveTypeAndDepth(t)
```
alexzielenski, line 527: Renamed `t->baseT` since `getNestedDefault` would always be invoked on the primitive base type like `string`, `int`, etc and always yield no nested default string. 

It seems the original intention of the code was to call `getNestedDefault` on the input argument. I've refactored `t->baseT` usages elsewhere where the base primitive element type was desired.

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -803,15 +804,40 @@ func (g *genDefaulter) GenerateType(c *generator.Context, t *types.Type, w io.Wr
 
 			// Rewrite the fully qualified name using the local package name
 			// from the imports
-			localPackage := g.imports.LocalNameOf(parsedName.Package)
-			if len(localPackage) > 0 {
-				current.defaultValue.SymbolReference = localPackage + "." + parsedName.Name
+			if len(parsedName.Package) > 0 && g.isOtherPackage(parsedName.Package) {
```
alexzielenski, line 807: Fixes a bug where fully specifying a symbol in local package would cause package name to be used.

alexzielenski (commented):
Annotated some changes to give context to reviewer

apelisse:
That's great, I still need to continue reviewing, I just want to mention that these defaults won't be included in the OpenAPI. The code that parses the default flags into their OpenAPI components lives in kube-openapi.

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -115,3 +120,72 @@ type DefaultedWithFunction struct {
...
+	// +default=ref(SomeValue)
+	AliasNonPointer ValueItem `json:",omitempty"`
+
+	// Can have a pointer to an alias whose default is a non-pointer value
+	// +default=ref(SomeValue)
+	AliasPointer *ValueItem `json:",omitempty"`
+
+	// Can have pointer to alias
+	// +default=ref(SomeDefault)
+	SymbolReference string
```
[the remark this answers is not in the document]
alexzielenski: Thats probably checked in kube-openapi. Should I remove the test?

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -115,3 +120,69 @@ type DefaultedWithFunction struct {
...
+
+	// Can override default specified on an alias
+	// +default=ref(SomeDefault)
+	AliasOverride Item
+
+	// Type-level default is not respected unless a pointer
+	AliasDefault DefaultedValueItem `json:",omitempty"`
+
+	// Type-level default is not respected unless a pointer
+	AliasPointerDefault *DefaultedValueItem `json:",omitempty"`
```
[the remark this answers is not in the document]
alexzielenski: That's what I would've expected but the source seems explicitly written to only read the default from the struct/alias type if the type is pointer (`depth > 0`):

https://github.com/kubernetes/gengo/blob/ab3349d207d4cc9f4bc5c37ba6aa062b4b2ffacc/examples/defaulter-gen/generators/defaulter.go#L518-L528

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -1029,15 +1055,44 @@ func (n *callNode) writeDefaulter(varName string, index string, isVarPointer boo
...
+
 			sw.Do(fmt.Sprintf("if %s == nil {\n", variablePlaceholder), args)
-			sw.Do("var ptrVar$.varDepth$ $.varType$ = $.defaultValue$\n", args)
-			// We iterate until a depth of 1 instead of 0 because the following line
-			// `if $.varName$ == &ptrVar1` accounts for 1 level already
-			for i := n.defaultDepth; i > 1; i-- {
-				sw.Do("ptrVar$.ptri$ := &ptrVar$.i$\n", generator.Args{"i": fmt.Sprintf("%d", i), "ptri": fmt.Sprintf("%d", (i - 1))})
+			if len(n.defaultValue.SymbolReference) == 0 {
+				// If default value is a literal then it can be assigned via var stmt
+				sw.Do(fmt.Sprintf("var ptrVar$.varDepth$ %s = $.defaultValue$\n", baseElementType), args)
```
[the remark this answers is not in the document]
alexzielenski: Not all variables were not available in the `args` map defined at the start of the function. The original source also use Sprint where necessary for this reason. Admittedly some spots I could've used the args. Will change those, but some Sprint calls will remain

examples/defaulter-gen/_output_tests/marker/type.go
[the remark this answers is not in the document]
apelisse: That seems somewhat unrelated, or at least testing two things at the same time in that case, I think it's fine to remove that test yeah.

examples/defaulter-gen/generators/defaulter.go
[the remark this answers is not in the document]
alexzielenski: there actually weren't any other spots to change to use `args`. All the `Sprint` formatters I used were contextual

examples/defaulter-gen/_output_tests/marker/type.go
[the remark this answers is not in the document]
alexzielenski: I just added `omitempty` json tag to make the test technically correct since id like to keep value type cases

examples/defaulter-gen/_output_tests/marker/type.go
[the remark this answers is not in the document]
apelisse: That's probably a bug to be honest.

examples/defaulter-gen/_output_tests/marker/type.go
```diff
@@ -115,3 +120,72 @@ type DefaultedWithFunction struct {
...
+
+	// Can have alias that is a pointer to type of constant
+	// +default=ref(SomeDefault)
+	AliasPointerInside Item
+
+	// Can override default specified on an alias
+	// +default=ref(SomeDefault)
+	AliasOverride Item
+
+	// Type-level default is not respected unless a pointer
```
apelisse, line 144: Not sure if that makes sense or not.

apelisse (commented):
Thanks
/lgtm

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -883,7 +885,8 @@ type callNode struct {
 	// markerOnly is true if the callNode exists solely to fill in a default value
 	markerOnly bool
 
-	// defaultDepth is used to determine pointer level of the default value
+	// defaultDepth is used to determine how to construct a pointer value
+	// for
```
apelisse, line 889: sentence is not finished?

apelisse (commented):
/lgtm

examples/defaulter-gen/generators/defaulter.go
alexzielenski: expanded comment

apelisse (commented):
/lgtm

apelisse:
`pull-gengo-test` seems to have errors but the job is still succeeding, is this expected?

alexzielenski:
> pull-gengo-test seems to have errors but the job is still succeeding, is this expected?
yes those are expected errors from the import boss test

alexzielenski:
Tested against current k8s master branch with only one minor package naming change (due to ignored local package no longer impacting naming):

```diff
diff --git a/pkg/controller/apis/config/v1alpha1/zz_generated.defaults.go b/pkg/controller/apis/config/v1alpha1/zz_generated.defaults.go
index bd27200c5a4..b441808f2ec 100644
--- a/pkg/controller/apis/config/v1alpha1/zz_generated.defaults.go
+++ b/pkg/controller/apis/config/v1alpha1/zz_generated.defaults.go
@@ -23,7 +23,7 @@ package v1alpha1
 
 import (
 	runtime "k8s.io/apimachinery/pkg/runtime"
-	cloudproviderconfigv1alpha1 "k8s.io/cloud-provider/config/v1alpha1"
+	configv1alpha1 "k8s.io/cloud-provider/config/v1alpha1"
 	v1alpha1 "k8s.io/kube-controller-manager/config/v1alpha1"
 )
 
@@ -39,5 +39,5 @@ func RegisterDefaults(scheme *runtime.Scheme) error {
 
 func SetObjectDefaults_KubeControllerManagerConfiguration(in *v1alpha1.KubeControllerManagerConfiguration) {
 	SetDefaults_KubeControllerManagerConfiguration(in)
-	cloudproviderconfigv1alpha1.SetDefaults_KubeCloudSharedConfiguration(&in.KubeCloudShared)
+	configv1alpha1.SetDefaults_KubeCloudSharedConfiguration(&in.KubeCloudShared)
 }
```

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -442,16 +442,16 @@ func newCallTreeForType(existingDefaulters, newDefaulters defaulterFuncMap) *cal
 	}
 }
 
-func resolveTypeAndDepth(t *types.Type) (*types.Type, int) {
+func resolveTypeAndDepth(t *types.Type) (*types.Type, []*types.Type) {
 	var prev *types.Type
-	depth := 0
+	var depth []*types.Type
```
sttts, line 447: this is not really depth anymore. What is it?

alexzielenski: Renamed `defaultDepth` to `pointerElementPath`, and separated this into a separate function `getPointerElementPath` with detailed comments

apelisse (commented):
/lgtm

PR #247, opened 2023-08-18
Export utils for import tracking and refs

alexzielenski:
Avoids duplication of default tag parsing to be used in OpenAPI-gen
* Exports the function used by `defaulter-gen` to parse symbols for use with OpenAPI. 
* Refactors the same function to take the source comment package path as input to avoid bugs with local symbols

Also includes some cleanup around the ImportTracker since OpenAPI-gen also needs to use `AddSymbol`:
* Adds `NewImportTrackerForPackage` so that local package can be provided to `DefaultImportTracker`
* Exposes `AddSymbol` on `ImportTracker` to be used by OpenAPI-gen and defaulter-gen
* Removes defaulter-gen's local `symbolTracker` interface since it is no longer necessary

required by https://github.com/kubernetes/kube-openapi/pull/419


generator/import_tracker.go
```diff
@@ -26,15 +26,18 @@ import (
 	"k8s.io/gengo/types"
 )
 
-func NewImportTracker(typesToAdd ...*types.Type) *namer.DefaultImportTracker {
-	tracker := namer.NewDefaultImportTracker(types.Name{})
+func NewImportTrackerForPackage(local string, typesToAdd ...*types.Type) *namer.DefaultImportTracker {
```
apelisse, line 29: We should probably have a doc?

apelisse (commented):
I don't see a test being updated, shouldn't we have a test if this becomes a public method?

examples/defaulter-gen/generators/defaulter.go
```diff
@@ -501,21 +501,27 @@ func mustEnforceDefault(t *types.Type, depth int, omitEmpty bool) (interface{},
 var refRE = regexp.MustCompile(`^ref\((?P<reference>[^"]+)\)$`)
 var refREIdentIndex = refRE.SubexpIndex("reference")
 
-// parseAsRef looks for strings that match one of the following:
+// ParseSymbolReference looks for strings that match one of the following:
```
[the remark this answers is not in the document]
alexzielenski: All the defaulter-gen source is under `examples`. I also think it is weird. I don't think I can move the whole package to its own location right now. Upstream k8s `code-generator` packages also  already refer to their public symbols from `examples` package such as: 

https://github.com/kubernetes/kubernetes/blob/33555943dc8d9a1a0c048d2ed0404106782f3f1c/staging/src/k8s.io/code-generator/cmd/deepcopy-gen/args/args.go#L24

https://github.com/kubernetes/kubernetes/blob/4d166947cf2b9d3420ae4ff1ff1da505486b74f5/staging/src/k8s.io/code-generator/cmd/prerelease-lifecycle-gen/prerelease-lifecycle-generators/status.go#L27

These packages are used as more than "examples"

generator/import_tracker.go
alexzielenski: added doc

alexzielenski:
added local package test

PR #245, opened 2023-08-17
be able to run defaulter-gen tests via `go test`

alexzielenski:
Continuation of #240 which went stale and closed.

> Files whose names begin with "_" are ignored by go test so none of the defaulter-gen tests were running in CI.
> [Some] tests failed because some packages missed a fake_deepcopy_conversion.go file. This is necessary because types registered in the scheme need to implement the DeepCopyObject() method.

Took into account some of the feedback provided on that PR, keeping `_output-tests` but adding its own go.mod so there can be a dependency on k/k packages directly without forming a cycle.

Changes from initial PR:
* Move `output_tests` back to `_output_tests`
* Added go.mod to _output_tests directory
* Rename package path from `_output_tests` to `output_tests` - otherwise retaining underscore in go package name caused ambiguous import error with gengo's file tree:
The module name itself still needed the underscore removed otherwise would get this error:
```console
ambiguous import: found package k8s.io/gengo/examples/defaulter-gen/_output_tests/empty in multiple modules:
        k8s.io/gengo v0.0.0 ($GOMOD/k8s.io/gengo@v0.0.0/examples/defaulter-gen/_output_tests/empty)
        k8s.io/gengo/examples/defaulter-gen/_output_tests (/Users/alex/go/src/k8s.io/gengo/examples/defaulter-gen/_output_tests/empty)
ambiguous import: found package k8s.io/gengo/examples/defaulter-gen/_output_tests/marker in multiple modules:
        k8s.io/gengo v0.0.0 ($GOMOD/k8s.io/gengo@v0.0.0/examples/defaulter-gen/_output_tests/marker)
        k8s.io/gengo/examples/defaulter-gen/_output_tests (/Users/alex/go/src/k8s.io/gengo/examples/defaulter-gen/_output_tests/marker)
ambiguous import: found package k8s.io/gengo/examples/defaulter-gen/_output_tests/marker/external in multiple modules:
        k8s.io/gengo v0.0.0 ($GOMOD/k8s.io/gengo@v0.0.0/examples/defaulter-gen/_output_tests/marker/external)
        k8s.io/gengo/examples/defaulter-gen/_output_tests (/Users/alex/go/src/k8s.io/gengo/examples/defaulter-gen/_output_tests/marker/external)
ambiguous import: found package k8s.io/gengo/examples/defaulter-gen/_output_tests/marker/external/external in multiple modules:
[...]

PR #248, opened 2023-08-23
Add go.work

alexzielenski:
* Adds a `go.work` now that `examples/defaulter-gen/_output_tests` is its own module.
* Renames `output_tests` module to `_output_tests` and remove disclaimer comment

This resolves `gopls` IDE errors, and also resolves the ambiguous import error discussed in #245 and allow the module to be named the same way as its source path.

/cc @apelisse


alexzielenski:
/close

I think this is a bad idea due to [go.work doc](https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#rationale-the-file)

> go.work files should not be checked into version control repos containing modules so that the go.work file in a module does not end up overriding the configuration a user created themselves outside of the module. The go.work documentation should contain clear warnings about this.

PR #249, opened 2023-08-23
Add verify-examples.sh and switch from makefiles to go test and go generate

alexzielenski:
This PR:

* Adds `generate.go` to deepcopy-gen and defaulter-gen, and fixes the `set-gen` generate to not rely on GOPATH. This enables the more standard `go generate` to be used over ad-hoc makefiles
* Adds `hack/verify-examples.sh` which will be called from CI to:
    1.  Remove generated code
    2. Re-generate code
    3. Run output tests
    4. Check for diff with git
* Changes most example tests to specify the boilerplate header path. Otherwise would rely on having k8s.io/gengo in your `GOPATH` to run
* Removes all makefiles in favor of `go generate` and `go test` workflows

After this PR we will be able to regenerate and rerun all examples tests in CI by invoking `hack/verify-examples.sh`

/cc @apelisse


apelisse:
Awesome, can't wait to see the CI!
/lgtm
/approve

alexzielenski:
/hold

still making this better

apelisse:
I don't know how I feel about having both the makefile and the shell script?

alexzielenski:
> I don't know how I feel about having both the makefile and the shell script?

agreed. removed the makefiles since they are supplanted by the verify script. However import-boss still retains a makefile since its test is quite different. I could refactor it to run as part of `go test` but it would have to just be a test that invokes `go run` of it self

PR #251, opened 2023-09-23
Defaults: Don't look at JSON omitempty

thockin:
Looking at JSON is weird, and the intersection of non-omitempty and having a default seems valid.

It seems like this was asserting that "if you have an out-of-bounds zero-value, you must set omitempty", but I don't buy that and looking at JSON tags is just weird.

### DIFF
--- a/examples/defaulter-gen/_output_tests/marker/fake_deepcopy_conversion.go
+++ b/examples/defaulter-gen/_output_tests/marker/fake_deepcopy_conversion.go
@@ -42,6 +42,27 @@ func (in *Defaulted) DeepCopyObject() runtime.Object {
 	return nil
 }
 
+func (in *DefaultedOmitempty) DeepCopy() *DefaultedOmitempty {
+	if in == nil {
+		return nil
+	}
+	out := new(DefaultedOmitempty)
+	in.DeepCopyInto(out)
+	return out
+}
+
+func (in *DefaultedOmitempty) DeepCopyInto(out *DefaultedOmitempty) {
+	*out = *in
+	return
+}
+
+func (in *DefaultedOmitempty) DeepCopyObject() runtime.Object {
+	if c := in.DeepCopy(); c != nil {
+		return c
+	}
+	return nil
+}
+
 func (in *DefaultedWithFunction) DeepCopy() *DefaultedWithFunction {
 	if in == nil {
 		return nil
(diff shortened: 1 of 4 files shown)


thockin:
hold for discussion

alexzielenski (approved):
/lgtm

agreed I cant think of a reason why `omitempty` should be forced into a default of zero. And for the purposes of defaulter-gen it makes sense to ignore zero-valued defaults (in constrast to OpenAPI-gen)

apelisse:
So we can continue discussing it before merging.
/hold

apelisse:
cc @liggitt

apelisse:
> agreed I cant think of a reason why omitempty should be forced into a default of zero.

Actually I read "why non-omitempty" should be forced into a default of zero. omitempty shouldn't have to.

apelisse:
> I think that is why putting a default on a non-omitempty field is weird, because clients are unaware 0 will be defaulted.

Yes, I think that's an important point, and also something they can not avoid when using typed clients.

apelisse:
> In [KEP-4153](https://github.com/alexzielenski/enhancements/blob/declarative-validation/keps/sig-api-machinery/4153-declarative-validation/README.md#zero-valued-primitive-fields) we discuss Zero-valued primitive fields and problems they cause for unstructured validation. The proposed solution there is to annotate x-kubernetes-strip-zero: true to non-pointer primitive fields to inform clients that the empty value is treated as unset. I think that would also clear the ambiguity the code generator restriction is trying to prevent here.

Yeah, the whole point of the exercise we've done in the defaulting KEP was to avoid introducing a new non-standard extension. Didn't we argue that we could just apply the defaults before running the CEL expression to have consistency between built-ins and CRDs?

thockin:
> all typed-clients are sending invalid values over the wire

Only if they send JSON.  And even so, who cares?  Most (all) of these fields ARE `omitempty` in k/k so this is a no-op except that it eliminates a weird corner case and layering violation.

Why shouldn't we also look at proto tags?  Or Yaml tags?

> I think that is why putting a default on a non-omitempty field is weird, because clients are unaware 0 will be defaulted.

In hindsight, yes and we have (mostly) stopped doing that :)

apelisse:
> Why shouldn't we also look at proto tags? Or Yaml tags?

Yes, OK, that's weird ;-)

thockin:
That's my point.  Defaulting code shouldn't be looking at JSON tags.

thockin:
I just realized this is still open - @alexzielenski or @jpbetz - opinions?

jpbetz:
It took me a while to figure out what was happening before and, yeah, I'm fully behind this approach.

Code looks good, I couldn't find anything to pick at.

/lgtm
/approve

Leaving the hold in your hands @thockin

thockin:
I verified that this does not change any generated code in k/k

PR #254, opened 2023-11-22
Stop using sets-gen in deepcopy-gen

skitt:
This will ultimately allow sets-gen to be removed.

### DIFF
--- a/examples/deepcopy-gen/generators/deepcopy.go
+++ b/examples/deepcopy-gen/generators/deepcopy.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	"k8s.io/gengo/args"
-	"k8s.io/gengo/examples/set-gen/sets"
 	"k8s.io/gengo/generator"
 	"k8s.io/gengo/namer"
 	"k8s.io/gengo/types"
@@ -131,7 +130,6 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 		klog.Fatalf("Failed loading boilerplate: %v", err)
 	}
 
-	inputs := sets.NewString(context.Inputs...)
 	packages := generator.Packages{}
 	header := append([]byte(fmt.Sprintf("//go:build !%s\n// +build !%s\n\n", arguments.GeneratedBuildTag, arguments.GeneratedBuildTag)), boilerplate...)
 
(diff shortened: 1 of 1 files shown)


examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -131,7 +131,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 		klog.Fatalf("Failed loading boilerplate: %v", err)
 	}
 
-	inputs := sets.NewString(context.Inputs...)
+	inputs := set.New(context.Inputs...)
```
thockin, line 134: Even simpler: don't need a set here at all.

```diff
@@ -24,7 +24,6 @@ import (
        "strings"
 
        "k8s.io/gengo/v2/args"
-       "k8s.io/gengo/v2/examples/set-gen/sets"
        "k8s.io/gengo/v2/generator"
        "k8s.io/gengo/v2/namer"
        "k8s.io/gengo/v2/types"
@@ -131,7 +130,6 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
                klog.Fatalf("Failed loading boilerplate: %v", err)
        }
 
-       inputs := sets.NewString(context.Inputs...)
        packages := generator.Packages{}
        header := append([]byte(fmt.Sprintf("//go:build !%s\n// +build !%s\n\n", arguments.GeneratedBuildTag, arguments.GeneratedBuildTag)), boilerplate...)
 
@@ -147,7 +145,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
                }
        }
 
-       for i := range inputs {
+       for _, i := range context.Inputs {
                klog.V(5).Infof("Considering pkg %q", i)
                pkg := context.Universe[i]
                if pkg == nil {
```

I made this change in my gengo/v2 PR, which I think is the right moment to remove set-gen.  As such, I think we should leave it alone in v1, or else make it the smallest possible PR.

skitt: Fair point, removing sets-gen requires a v2 anyway. I’ve updated the PR to only drop the use of sets-gen — I still think this is useful in v1 to reduce the number of generated deepcopies using sets-gen.

thockin (commented):
/lgtm
/approve

Issue #225, opened 2022-06-06
[should gengo add] Support for 1.18 Generics


austince:
When experimenting with using go 1.18 generics in some k8s type definitions, I'm hitting errors (as one would expect with such a large language change).  I'm running conversion, defaulter, and deepcopy generation on the following types. 

```go

type ListItems[T any] struct {
	Items []T `json:"items"`
}

// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object

type TestObjectList struct {
	metav1.TypeMeta              `json:",inline"`
	metav1.ListMeta              `json:"metadata"`
	metav1.ListItems[TestObject] `json:",inline"`
}

// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object

type TestObject struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata"`
}
```

The errors seem to start with the gengo parser not having a case for the generic type param: https://github.com/kubernetes/gengo/blob/4627b89bbf1bc415e134f359f597053a5d1b6553/parser/parse.go#L857-L865

```
conversion generation
W0606 16:15:58.292211  133284 parse.go:862] Making unsupported type entry "T" for: &types.TypeParam{check:(*types.Checker)(nil), id:0x1, obj:(*types.TypeName)(0xc008bb1630), index:0, bound:(*types.Interface)(0xc0000942a0)}

defaulter generation
W0606 16:16:03.505939  135029 parse.go:862] Making unsupported type entry "T" for: &types.TypeParam{check:(*types.Checker)(nil), id:0x1, obj:(*types.TypeName)(0xc009fa7e50), index:0, bound:(*types.Interface)(0xc00007e2a0)}

deepcopy generation
W0606 16:16:14.790210  136792 parse.go:862] Making unsupported type entry "T" for: &types.TypeParam{check:(*types.Checker)(nil), id:0x1, obj:(*types.TypeName)(0xc00b6160f0), index:0, bound:(*types.Interface)(0xc00007e2a0)}
F0606 16:16:14.806387  136792 deepcopy.go:816] Hit an unsupported type T for []T
goroutine 1 [running]:
k8s.io/klog/v2.stacks(0x1)
        /home/austin/go/pkg/mod/k8s.io/klog/v2@v2.60.1/klog.go:860 +0x8a
[...]

PR #255, opened 2023-11-28
Support parsing type parameters

TheSpiritXIII:
This PR allows the parser to parse generics (partial fix for #225).

However, this PR **does not** allow generators to take advantage of it. For example, the deepcopy generator will still fail if you use a generic field, because it does not know whether that generic has a `DeepCopy` method. Although, the deepcopy generator will now work when you use a phantom type, which was not possible before. What are the use-cases for a phantom type? No clue, but I figured splitting the fix for #225 into multiple PRs would be easier to get reviewed and merged.

### DIFF
--- a/v2/parser/parse.go
+++ b/v2/parser/parse.go
@@ -587,9 +587,16 @@ func goNameToName(in string) types.Name {
 		return types.Name{Name: in}
 	}
 
+	// There may be '.' characters within a generic. Temporarily remove
+	// the generic.
+	genericIndex := strings.IndexRune(in, '[')
+	if genericIndex == -1 {
+		genericIndex = len(in)
+	}
+
 	// Otherwise, if there are '.' characters present, the name has a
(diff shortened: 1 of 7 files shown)


thockin:
If you are still keen to merge this - could you rebase it to gengo/v2?  I'd like to see this fixed.

TheSpiritXIII:
> If you are still keen to merge this - could you rebase it to gengo/v2? I'd like to see this fixed.

Thanks! I've split this PR into two, as deepcopy has moved into kubernetes/kubernetes. The deepcopy test PR can be found here: https://github.com/kubernetes/kubernetes/pull/123996 Note, I had to upgrade `gengo` in that repository. Also, since Go workspaces don't support local replace directives, I used my local copy for now and marked the PR as WIP until this PR is released.

thockin (commented):
Thanks!

/lgtm
/approve

PR #256, opened 2023-11-29
Support generating a DeepCopy for struct generics which implement DeepCopy

TheSpiritXIII:
Part 2 of #255, adding support for self-referential generics to implement DeepCopy on generic parameter types, which is then used by the object DeepCopy method.

This also provides a solution for #225. An alternative or supplement to this feature would be to support generating `DeepCopyX` where `X` is the generic parameter name. However, this PR's solution is arguably easier for users to maintain and what most users would want.


thockin:
Closing in favor of v2

PR #257, opened 2023-12-22
Support parsing directives in comment of methods, structs, struct members and package doc

longit644:
This PR add support for parsing directives in comment of methods, structs, and struct members

### DIFF
--- a/parser/local_parse_test.go
+++ b/parser/local_parse_test.go
@@ -17,8 +17,11 @@ limitations under the License.
 package parser
 
 import (
+	"go/ast"
 	"os"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 )
 
 func TestImportBuildPackage(t *testing.T) {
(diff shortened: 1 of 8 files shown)


thockin:
Hi.  I am currently overhauling gnengo (https://github.com/kubernetes/gengo/pull/259) and looking at PRs.  

This PR doesn't explain WHY you need this, and keeping a fork of ast is not something we really want to do.

IF we were to consider this it would be on v2, and I don't think we want to at first glance.

longit644:
Hi @thockin, 

Thank you for your response.

This PR adds parsing directive to support another new marker creation method for generator. It uses directive syntax `//k8s:...`, similar to how recent Go versions have transitioned from `// +build` to `//go:build`

If you believe this adds value to `gengo`, I would be happy to create another PR on `v2`.

thockin:
Do those not get retained as comments?

I don't think we want to keep a fork of AST for this, especially after we move to packages.Load (PR pending)

longit644:
No, directives are not retained within the comments. They are filtered out during the parsing process. 

As for keeping the fork of AST, I agree that may not be the most efficient approach. Instead, we could consider re-implementing the directive check or exploring alternative solutions.

thockin:
I don't object to the goal, but I think it's trickier with `packages.Load` - try from my v2 PR as a base?

PR #259, opened 2024-01-28
Spawn gengo/v2 (KEP 4402)

thockin:
First: This PR can be broken up!

See https://github.com/kubernetes/enhancements/issues/4402

This clones gengo into gengo/v2 and overhauls it to use Go's `packages.Load()` instead of DIY.  It includes a number of refactorings and simplifcations, too, which really warrant being "v2".

The "examples" (which were actually prod tools) get moved to k/k and replaced with simpler examples.

The commits in this PR were adapted from https://github.com/kubernetes/kubernetes/pull/122624

I have tried to make sure test coverage did not drop and that tests pass at every commit.


v2/generator/execute.go
```diff
@@ -0,0 +1,266 @@
...
+		c2.Namers[k] = v
+	}
+
+	for name, namer := range namers {
+		c2.Namers[name] = namer
+	}
+	return &c2
+}
+
+// ExecuteTargets runs the generators for a single target.
```
skitt, line 166:
```suggestion
// ExecuteTarget runs the generators for a single target.
```

v2/generator/execute.go
```diff
@@ -207,26 +203,11 @@ func (c *Context) addNameSystems(namers namer.NameSystems) *Context {
...
-	// The following will trim a path prefix (github.com/foo/bar) from p.Path() to arrive at
-	// a relative path that works with projects not in GOPATH.
-	if c.TrimPathPrefix != "" {
-		separator := string(filepath.Separator)
-		if !strings.HasSuffix(c.TrimPathPrefix, separator) {
-			c.TrimPathPrefix += separator
-		}
-
-		path = strings.TrimPrefix(path, c.TrimPathPrefix)
+// ExecutePackages runs the generators for a single package.
```
skitt:
```suggestion
// ExecutePackage runs the generators for a single target.
```

v2/parser/parse.go
```diff
@@ -42,43 +38,32 @@ type importPathString string
...
+	goPkgs map[importPathString]*packages.Package
 
-	// Set by typeCheckPackage(), used by importPackage() and friends.
-	typeCheckedPackages map[importPathString]*tc.Package
-
-	// Map of package path to whether the user requested it or it was from
-	// an import.
+	// Keep track of which packages were directly requested (as opposed to
+	// those which are transitively loaded).
 	userRequested map[importPathString]bool
```
skitt: How about using sets here? It would simplify `UserRequestedPackages()` etc. Or is it desirable to avoid a dependency on `k8s.io/utils` in gengo?

v2/parser/parse.go
```diff
@@ -42,43 +38,32 @@ type importPathString string
...
 
-	// All comments from everywhere in every parsed file.
-	endLineToCommentGroup map[fileLine]*ast.CommentGroup
+	// Keep track of which packages have already been scanned for types.
+	fullyProcessed map[importPathString]bool
```
skitt: Same as above.

v2/generator/execute.go
```diff
@@ -210,9 +210,16 @@ func (c *Context) ExecutePackage(p Package) error {
 		return fmt.Errorf("no source-path for package %s", p.Path())
 	}
 	klog.V(5).Infof("Processing package %q, disk location %q", p.Name(), path)
+
 	// Filter out any types the *package* doesn't care about.
 	packageContext := c.filteredBy(p.Filter)
-	os.MkdirAll(path, 0755)
+
+	if !c.Verify {
+		if err := os.MkdirAll(path, 0755); err != nil {
```
skitt: This is stylistic preference, but I like `0o755` to make octal values really explicit (same as hex and binary).

v2/args/args.go
```diff
@@ -137,7 +137,7 @@ func (g *GeneratorArgs) NewBuilder() (*parser.Builder, error) {
 // Execute implements main().
 // If you don't need any non-default behavior, use as:
 // args.Default().Execute(...)
-func (g *GeneratorArgs) Execute(nameSystems namer.NameSystems, defaultSystem string, pkgs func(*generator.Context, *GeneratorArgs) []generator.Package) error {
+func (g *GeneratorArgs) Execute(nameSystems namer.NameSystems, defaultSystem string, getTargets func(*generator.Context, *GeneratorArgs) []generator.Target) error {
```
skitt: If the goal is to distinguish input and output packages, how about making that explicit in the type names? `InputPackage` / `OutputPackage`, or `InPackage` / `OutPackage`, or `InputPkg` etc.

v2/generator/execute.go
thockin: ACK

v2/parser/parse.go
thockin: I'd avoid the dep - "a little copy is better than a little dep" :)

thockin: Also could make vendoring into k/k more messy

v2/generator/execute.go
thockin: I didn't even know that was a thing.  IMO "0755" is UNIX and C since...forever.  Everyone knows what that means.

thockin:
Fixed minor feedback (thanks!)

v2/args/args.go
skitt: I was reviewing this in the context of https://github.com/kubernetes/gengo/pull/259/commits/da029b0c88feccc77a402942cda0e7782a97dbdd — the commit message says

> Having two types names "Package" was super confusing to read.  I don't love "Target" but at least it implies "this is the output" rather than the input.

Since the use of `Package` is ambiguous, it seems to me that a possible solution would be to disambiguate both uses cases instead of using a not-great replacement for the output scenario; `InputPackage` / `OutputPackage` for example.

thockin: Oh, I see.  I was shooting for something shortish rather than renaming `types.Package` at this time.  In truth, I rather hate the way gengo has its own `Package` type with (IMO) dangerous lazy-load semantics and how it interposes its own `Type` and `Kind` types with different semantics from Go.  I'd like to see use eliminate all of that, but it wasn't necessary to this PR and I *really* want to get workspaces support into k/k (which is blocked on this :)

So - we could totally do a followup to rename things more :)

skitt: That’s fine by me!

v2/hack/verify-examples.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+# hack/verify-examples.sh
```
jpbetz, line 3: Just thinking ahead for a moment..

We should tool our testing infra to run v2/hack/verify-examples.sh? We could run it from the root `hack/verify-examples.sh` but that running it separately will probably result in clearer error reporting.

v2/README.md
```diff
@@ -0,0 +1,42 @@
+# Gengo: a framework for building simple code generators
```
jpbetz, line 1: Do we need to update the root gengo readme to say anything about v2 (i.e. https://github.com/kubernetes/gengo/blob/master/README.md)?

jpbetz, line 1: Can we get an OWNERS file added here in the v2 directory?

v2/comments.go
```diff
@@ -0,0 +1,83 @@
...
+		// TODO: we could support multiple values per key if we split on spaces
+		kv := strings.SplitN(line[len(marker):], "=", 2)
+		if len(kv) == 2 {
+			out[kv[0]] = append(out[kv[0]], kv[1])
+		} else if len(kv) == 1 {
+			out[kv[0]] = append(out[kv[0]], "")
+		}
+	}
+	return out
+}
```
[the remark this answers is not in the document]
alexzielenski: I'm not sure we will since a key difference between the two is the kube-openapi version does not allow implicit arrays by duplicating markers which I dont think would be backwards compatible if we merged the two (unless it was a parser optoin or something)

v2/README.md
```diff
@@ -0,0 +1,42 @@
...
+[simple examples](./examples) in this repo or the more extensive tools in the
+Kubernetes [code-generator](https://github.com/kubernetes/code-generator/)
+repo.
+
+## Overview
+
+Gengo is used to build tools (generally a tool is a binary).  Each tool
+describes some number of `Targets`. A target is a single output package, which
+may be the same as the inputs (if the tool generates code alongside the inputs)
+or different.  Each `Target` describes some number of `Generators`.  A
```
alexzielenski, line 32: +1 to this straightforward API separation

v2/generator/import_tracker.go
```diff
@@ -0,0 +1,89 @@
...
+//	tracker := NewImportTrackerForPackage("bar.com/pkg/foo")
+//	tracker.AddSymbol(types.Name{"bar.com/pkg/foo.MyType"})
+//	tracker.AddSymbol(types.Name{"bar.com/pkg/baz.MyType"})
+//	tracker.AddSymbol(types.Name{"bar.com/pkg/baz/baz.MyType"})
+//
+//	tracker.LocalNameOf("bar.com/pkg/foo") -> ""
+//	tracker.LocalNameOf("bar.com/pkg/baz") -> "baz"
+//	tracker.LocalNameOf("bar.com/pkg/baz/baz") -> "bazbaz"
+//	tracker.ImportLines() -> {`baz "bar.com/pkg/baz"`, `bazbaz "bar.com/pkg/baz/baz"`}
+func NewImportTrackerForPackage(local string, typesToAdd ...*types.Type) *namer.DefaultImportTracker {
```
alexzielenski, line 45: I didn't see this used in any of the examples, is it still necessary?

v2/examples/kilroy/main.go
```diff
@@ -0,0 +1,209 @@
...
+	return t.Kind == types.Struct && !namer.IsPrivateGoName(t.Name.Name)
+}
+
+// Namers returns a set of NameSystems which will be merged with the namers
+// provided when executing this package. In case of a name collision, the
+// values produced here will win.
+func (g *kilroyGenerator) Namers(*generator.Context) namer.NameSystems {
+	return namer.NameSystems{
+		// This elides package names when the name is in "this" package.
+		"raw": namer.NewRawNamer(g.myPackage.Path, nil),
```
alexzielenski, line 182: Maybe this works the same as previous gengo but I'm noticing this shadows the `raw` name passed via `getNameSystems` to `gengo.Execute`. How do the two namesystems interact and handle conflicts?

v2/examples/pointuh/main.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+// Namers returns a set of NameSystems which will be merged with the namers
+// provided when executing this package. In case of a name collision, the
+// values produced here will win.
+func (g *pointuhGenerator) Namers(*generator.Context) namer.NameSystems {
+	return namer.NameSystems{
+		// This elides package names when the name is in "this" package.
+		"localraw": namer.NewRawNamer(g.myPackage.Path, nil),
```
alexzielenski, line 183: I've seen this pattern in a few examples. Is this something that could make a sensible default? I feel like the local-aware namer and per-target import trackers are something consumers of this library wouldnt mind being plumbed for them?

v2/comments.go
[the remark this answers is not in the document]
thockin: It would be good to have one convention that applied to all tools.  FWIW I didn't change this logic, just copied it over. :)

v2/hack/verify-examples.sh
thockin: Good point!

v2/README.md
thockin: good idea, will do

thockin: I assumed root was OK, but can copy that over if you want to make it different?  But do we want it to be different?  I don't think so.

v2/generator/simple_package.go
```diff
@@ -0,0 +1,77 @@
...
+
+package generator
+
+import (
+	"k8s.io/gengo/v2/types"
+)
+
+// SimpleTarget is implements Target in terms of static configuration.
+// The package name, path, and dir are required to be non-empty.
+type SimpleTarget struct {
```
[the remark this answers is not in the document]
thockin: good call!  In next push

v2/generator/import_tracker.go
thockin: It is still used by defaulter, even though it was moved to k/k

v2/examples/pointuh/main.go
thockin: I do think we could simplify how namers work more - I didn't cover it in this PR because it was already too big.

thockin:
pushed with some comments addressed

v2/examples/kilroy/main.go
thockin: I took a poke at this and realized that some of the tools use the Target-scope namers to avoid allocating the exact same namers for each Generator.  We could pass a `NameSystems` and a `Namer` into `Execute` (instead of a `NameSystems` and a `string`) but it doesn't save anything.

thockin:
OK.  KEPs are done.  I know this will take time, so I'm sending my first ping for review :)

alexzielenski (commented):
lgtm

I think we all agree on general approach of introducing gengo/v2. I can't find anything blocking on the few passes I've made. Since this is net-new change, with working prototypes and no comments for a while, should we merge this and fix any issues that come up as we use it?

v2/examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -297,9 +297,12 @@ func (g *genDeepCopy) copyableAndInBounds(t *types.Type) bool {
 // if the type does not match. This allows more efficient deep copy
 // implementations to be defined by the type's author.  The correct signature
 // for a type T is:
-//    func (t T) DeepCopy() T
```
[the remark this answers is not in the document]
thockin: I will hack a top-level `make test` as a new commit.

v2/examples/deepcopy-gen/generators/deepcopy.go
```diff
@@ -147,7 +145,7 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
 		}
 	}
 
-	for i := range inputs {
+	for _, i := range context.Inputs {
```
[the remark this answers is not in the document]
thockin: You mean to protect it from mutation?  If we want to be paranoid, we should turn it into a methods that returns a copy or something like that.

v2/hack/verify-examples.sh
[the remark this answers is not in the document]
thockin: cute.  Will do

v2/examples/defaulter-gen/generators/defaulter.go
```diff
@@ -243,15 +243,74 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
...
+		if len(inputTags) > 1 {
+			panic(fmt.Sprintf("there may only be one input tag, got %#v", inputTags))
+		}
+		if len(inputTags) == 1 {
+			inputPath := inputTags[0]
+			if strings.HasPrefix(inputPath, "./") || strings.HasPrefix(inputPath, "../") {
+				// this is a relative dir, which will not work under gomodules.
+				// join with the local package path, but warn
+				klog.Warningf("relative path %s=%s will not work under gomodule mode; use full package path (as used by 'import') instead", inputTagName, inputPath)
+				inputPath = filepath.Join(pkg.Path, inputTags[0])
```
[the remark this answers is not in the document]
thockin: Interesting distinction I didn't consider.  It's become habit to always use `filepath`.  I'm not sure these tools run on Windows to know if this breaks or if `path` is better.

v2/examples/defaulter-gen/generators/defaulter.go
```diff
@@ -243,15 +243,74 @@ func Packages(context *generator.Context, arguments *args.GeneratorArgs) generat
...
+			pkgToInput[i] = i
+		}
+	}
+
+	// Make sure explicit peer-packages are added.
+	var peerPkgs []string
+	if customArgs, ok := arguments.CustomArgs.(*CustomArgs); ok {
+		for _, pkg := range customArgs.ExtraPeerDirs {
+			if i := strings.Index(pkg, "/vendor/"); i != -1 {
+				pkg = pkg[i+len("/vendor/"):]
```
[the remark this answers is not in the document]
thockin: To hit this they would have to have said `./vendor/example.com/project` and I think that `packages.Load` would still DTRT.

README.md
```diff
@@ -9,6 +9,8 @@
 [GoReport]: https://goreportcard.com/badge/github.com/kubernetes/gengo
 [GoReportStatus]: https://goreportcard.com/report/github.com/kubernetes/gengo
 
+NOTE: [`gengo/2`](https://github.com/thockin/gengo/tree/master/v2) is the current development module.
```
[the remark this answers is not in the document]
thockin: nope!  Good catch.

v2/generator/golang_generator.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"io"
+
+	"k8s.io/gengo/v2/namer"
+	"k8s.io/gengo/v2/types"
+)
+
+const (
+	GolangFileType = "golang"
```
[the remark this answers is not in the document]
thockin: added a new commit

v2/examples/deepcopy-gen/generators/deepcopy.go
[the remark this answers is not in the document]
liggitt: No, I just didn't think we wanted to handle the same identical input twice

v2/examples/kilroy/main.go
```diff
@@ -60,33 +60,34 @@ func main() {
...
 	stdArgs := args.Default().WithoutDefaultFlagParsing()
-	stdArgs.OutputFileBaseName = "kilroy_generated.go"
 	toolArgs := &toolArgs{}
 	stdArgs.CustomArgs = toolArgs
 	return stdArgs, toolArgs
 }
 
 // AddFlags adds this tool's flags to the flagset.
 func (ta *toolArgs) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&ta.outputFile, "output-file", "kilroy_generated.go",
```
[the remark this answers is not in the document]
thockin: in next push

v2/examples/pointuh/main.go
```diff
@@ -64,41 +64,42 @@ func main() {
...
 
 // AddFlags add the generator flags to the flag set.
 func (ta *toolArgs) AddFlags(fs *pflag.FlagSet) {
-	fs.StringVar(&ta.OutputDir, "output-dir", "",
+	fs.StringVar(&ta.outputDir, "output-dir", "",
 		"the base directory under which to generate results")
-	fs.StringVar(&ta.OutputPkg, "output-pkg", "",
+	fs.StringVar(&ta.outputPkg, "output-pkg", "",
 		"the base Go import-path under which to generate results")
+	fs.StringVar(&ta.outputFile, "output-file", "pointuh_generated.go",
```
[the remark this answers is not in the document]
thockin: in next push

v2/args/args.go
```diff
@@ -49,23 +47,9 @@ type GeneratorArgs struct {
...
-	GeneratedByCommentTemplate string
-
 	// If true, include *_test.go files
 	IncludeTestFiles bool
 
-	// GeneratedBuildTag is the tag used to identify code generated by execution
-	// of this type. Each generator should use a different tag, and different
-	// groups of generators (external API that depends on Kube generations) should
-	// keep tags distinct as well.
-	GeneratedBuildTag string
```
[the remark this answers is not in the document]
thockin: I don't think we ever followed this for ourselves

v2/parser/parse.go
```diff
@@ -0,0 +1,780 @@
...
+}
+
+// key type for finding comments.
+type fileLine struct {
+	file string
+	line int
+}
+
+// New constructs a new builder.
+func New(buildTags []string) *Parser {
```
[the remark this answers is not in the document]
thockin: > that would let almost all existing New() invocations remain unchanged

I see exactly 2 calls - one which needs to pass the build tags and one in a test.  I can change it as you suggest, I just want to point out the limited utility :)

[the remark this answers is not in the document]
thockin: in next push
(OK, it's a bit cleaner :)

v2/generator/simple_target.go
```diff
@@ -0,0 +1,77 @@
...
+type SimpleTarget struct {
+	// PkgName is the name of the resulting package (as in "package xxxx").
+	// Required.
+	PkgName string
+	// PkgPath is the canonical Go import-path of the resulting package (as in
+	// "import example.com/xxxx/yyyy"). Required.
+	PkgPath string
+	// PkgDir is the location of the resulting package on disk (which may not
+	// exist yet). Required.
+	PkgDir string
```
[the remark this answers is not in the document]
thockin: in next push

v2/parser/parse.go
```diff
@@ -429,27 +429,26 @@ func (b *Builder) addPkgToUniverse(pkg *packages.Package, u *types.Universe) err
...
+		case *tc.Var:
+			if !obj.IsField() {
+				t := b.addVariable(*u, nil, obj)
+				b.addCommentsToType(obj, t)
+			}
+		case *tc.Const:
+			t := b.addConstant(*u, nil, obj)
 			b.addCommentsToType(obj, t)
+		default:
+			klog.Infof("addPkgToUniverse %q: unhandled object: %v", pkgPath, obj)
```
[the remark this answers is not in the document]
thockin: in next push

thockin:
Pushed with comments addressed (mostly new commits, but some fixed in-line)

thockin:
It's easier for me to merge this and then fix any ripple into the k/k and openapi PRs

PR #261, opened 2024-02-25
Allow tests to seed a packages.Config

thockin:
This is needed for tests which use `packagestest`, so it can inject env vars and change the working dir.  kube-openapi needs this.

### DIFF
--- a/v2/parser/parse.go
+++ b/v2/parser/parse.go
@@ -93,6 +93,11 @@ type Options struct {
 // FindPackages expands the provided patterns into a list of Go import-paths,
 // much like `go list -find`.
 func (p *Parser) FindPackages(patterns ...string) ([]string, error) {
+	return p.findPackages(nil, patterns...)
+}
+
+// baseCfg is an optional (may be nil) config which might be injected by tests.
+func (p *Parser) findPackages(baseCfg *packages.Config, patterns ...string) ([]string, error) {
 	toFind := make([]string, 0, len(patterns))
 	results := make([]string, 0, len(patterns))
 	for _, pat := range patterns {
@@ -106,11 +111,14 @@ func (p *Parser) FindPackages(patterns ...string) ([]string, error) {
 		return results, nil
 	}
 
-	cfg := packages.Config{
-		Mode:       packages.NeedName | packages.NeedFiles,
-		BuildFlags: []string{"-tags", strings.Join(p.buildTags, ",")},
-		Tests:      false,
+	cfg := packages.Config{}
+	if baseCfg != nil { // This is to support tests, e.g. to inject a fake GOPATH or CWD.
+		cfg = *baseCfg
 	}
+	cfg.Mode |= packages.NeedName | packages.NeedFiles
+	cfg.BuildFlags = append(cfg.BuildFlags, "-tags", strings.Join(p.buildTags, ","))
+	cfg.Tests = false
+
 	pkgs, err := packages.Load(&cfg, toFind...)
 	if err != nil {
 		return nil, fmt.Errorf("error loading packages: %w", err)
@@ -143,6 +151,13 @@ func (p *Parser) LoadPackages(patterns ...string) error {
 	return err
 }
 
+// LoadPackagesWithConfig loads and parses the specified Go packages with the
+// specified packages.Config as a starting point.  This is for testing.
+func (p *Parser) LoadPackagesWithConfig(cfg *packages.Config, patterns ...string) error {
+	_, err := p.loadPackagesWithConfig(cfg, patterns...)
+	return err
+}
+
 // LoadPackagesTo loads and parses the specified Go packages, and inserts them
 // into the specified Universe. It returns the packages which match the
 // patterns, but loads all packages and their imports, recursively, into the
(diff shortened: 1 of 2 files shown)


v2/parser/parse.go
```diff
@@ -106,11 +111,14 @@ func (p *Parser) FindPackages(patterns ...string) ([]string, error) {
...
-		Mode:       packages.NeedName | packages.NeedFiles,
-		BuildFlags: []string{"-tags", strings.Join(p.buildTags, ",")},
-		Tests:      false,
+	cfg := packages.Config{}
+	if baseCfg != nil { // This is to support tests, e.g. to inject a fake GOPATH or CWD.
+		cfg = *baseCfg
 	}
+	cfg.Mode |= packages.NeedName | packages.NeedFiles
+	cfg.BuildFlags = append(cfg.BuildFlags, "-tags", strings.Join(p.buildTags, ","))
+	cfg.Tests = false
```
aojea, line 120: should this not be into an else statement then?
you re overriding the config passed

thockin: Only overriding fields that we know we have to set, leaving other fields (like Dir and Env, which are what I need for the openapi tests to pass)

aojea: I don't know how nitpicking is this as I´m not familiar with this codebase, but sounds that you could achieve the same just making `baseCfg *packages.Config` part of the Parser struct , since it seems you'll always need to access the private methods, and you´ll avoid to modify the methods

v2/parser/parse.go
```diff
@@ -143,6 +151,13 @@ func (p *Parser) LoadPackages(patterns ...string) error {
 	return err
 }
 
+// LoadPackagesWithConfig loads and parses the specified Go packages with the
+// specified packages.Config as a starting point.  This is for testing.
```
liggitt, line 155: "This is for testing" seems unlikely to hold if this allows injecting tricky things that impact the package load...

The surface area of the injected cfg is large and definitely allows stuff that would break us...
* some of which we stomp (Tests=false, Fset = ...)
* some of which we aggregate in ways I'm not sure will work (appending possibly more build tags, combining additional things with mode)
* some of which we don't do anything with in ways I'm not sure will work (allowing esoteric options like ParseFile / Overlay to rule)

Is there a way to limit the surface area of what we can inject this way to *just* things we know won't conflict with or break our internal use (e.g. Dir / Env)?

```go
if cfg != nil {
  cfgCopy := *cfg
  cfgCopy.Dir = ""
  cfgCopy.Env = nil
  if cfgCopy != (Config{}) {
    return fmt.Errorf("only Dir and Env may be set in the base config, got additional fields %#v", cfgCopy)
  }
}
```

v2/parser/parse.go
liggitt: I'd _really_ like to keep the override as constrained as possible until we have good reasons to expand and can reason about the interactions with the internal optionis we set... could keep the same signature (pass in a base config), but limit it to Dir and Env until we have reason to expand that, and only use Dir and Env here and in `loadPackages`, e.g.:

```diff
+if baseCfg == nil {
+ 		baseCfg = &packages.Config{}
+}
 cfg := packages.Config{
 		Mode:       packages.NeedName | packages.NeedFiles,
 		BuildFlags: []string{"-tags", strings.Join(p.buildTags, ",")},
 		Tests:      false,
+		Dir:        baseCfg.Dir,
+		Env:        baseCfg.Env, 
 }
```

thockin: That works, too

v2/parser/parse.go
thockin: I could just take Dir and Env as args, or as you suggest above just extract those 2 fields from Config

thockin:
Re-pushed with change to ONLY extract Dir and Env

v2/parser/parse.go
```diff
@@ -143,6 +154,14 @@ func (p *Parser) LoadPackages(patterns ...string) error {
 	return err
 }
 
+// LoadPackagesWithConfig loads and parses the specified Go packages with the
+// specified packages.Config as a starting point.  This is for testing, and
+// only the .Dir and .Env fields of the Config will be considered.
```
liggitt, line 159: can you add a guard that only Dir and Env and error for now if anything else is set, to keep from accumulating usage that sets other fields and may break in the future if we decide to honor additional fields?

thockin: Unfortunately go's `packagestest` sets other fields, which would break this check:

```
        Config: &packages.Config{    
            Dir:     temp,    
            Env:     append(os.Environ(), "GOPACKAGESDRIVER=off", "GOROOT="), // Clear GOROOT to work around #32849.     
            Overlay: make(map[string][]byte),    
            Tests:   true,    
            Mode:    packages.LoadImports,    
        }, 
```

So a) it's good fortune that I was already setting `Tests: false`; b) it doesn't seem to need all of LoadImports, and works with our own `Mode`.

If you want to be super-paranoid, I can take `dir string, env []string` instead of `cfg *packages.Config` in the public API, and convert that to a smaller struct internally.  It just feels like a lot of work :)

Or I can rename the public function like `LoadPackagesWithConfigForTesting` and rely on the fact that this is not REALLY a public project anymore.

liggitt: eh, that's probably good enough

/lgtm
/approve

Issue #264, opened 2024-02-28
Add presubmit check to prevent accidental `go` directive bump


liggitt:
Bumping the go directive in the go.mod files propagates to dependents virally, including kube-openapi.

Since this repo only releases from HEAD, we should be intentional when updating this directive since we may have to update on older kubernetes release branches.

Until https://github.com/golang/go/issues/65573 is resolved and available, I'd suggest keeping the go directive in this repo no higher than the version of go reflected in the root `go.mod` file of the oldest supported kubernetes/kubernetes release branch

cc @thockin

PR #267, opened 2024-03-12
.*: add verify script for go directive changes

MadhavJivrajani:
Building on https://github.com/kubernetes/utils/pull/306 

This commit adds a script that checks the changed version in the go.mod file with a certain maximum version that the go directive can have.

We set the maximum version of the go directive as 1.20 here because the oldest go directive that exists on our supported release branches in k/k is 1.20.

Fixes https://github.com/kubernetes/gengo/issues/264 
xref https://github.com/kubernetes/kubernetes/issues/123744

~TODO: need to change the command here: https://github.com/kubernetes/test-infra/blob/0100e558a4a3594165262e0864a53d1b22f0973c/config/jobs/kubernetes/gengo/gengo-config.yaml#L29~

Done https://github.com/kubernetes/test-infra/pull/32368

/assign @liggitt @dims

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,12 @@ all:
 
 test:
 	go test -race ./...
-	./hack/verify-examples.sh
 	go -C v2 test -race ./...
 	(cd v2 && ./hack/verify-examples.sh)
+
+# We verify for the maximum version of the go directive as 1.20
+# here because the oldest go directive that exists on our supported
+# release branches in k/k is 1.20.
+verify:
+	./hack/verify-examples.sh
+	./hack/verify-go-directive.sh 1.20
--- a/hack/verify-go-directive.sh
+++ b/hack/verify-go-directive.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+function usage {
+  local script="$(basename $0)"
+
+  echo >&2 "Usage: ${script} <maximum go directive>
+This script should be run at the root of a module.
+
+Compare the go directive in the local working copy's go.mod
+to the specified maximum version it can be. Versions provided
+here are of the form 1.x.y, without the 'go' prefix.
+Examples:
+  ${script} 1.20
+  ${script} 1.21.6
+"
+  exit 1
+}
+
+max="$1"
+# If max is empty, print usage and error
+if [[ -z "${max}" ]]; then
+  usage;
+fi
+
+# Don't specify the version with the go prefix, just 1.x.y will do.
+if [[ ! "${max}" =~ ^[0-9]\.[0-9]+(\.[0-9]+)?$ ]]; then
+    usage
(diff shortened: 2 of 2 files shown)


hack/verify-go-directive.sh
```diff
@@ -0,0 +1,64 @@
...
+Examples:
+  ${script} -g 1.20
+  ${script} -g 1.21.6
+"
+  exit 1
+}
+
+max=""
+while getopts g: o
+do case "$o" in
```
thockin, line 39: Why not just `$script <version>` - do we need flag handling?

hack/verify-go-directive.sh
```diff
@@ -0,0 +1,64 @@
...
+  esac
+done
+
+# If max is empty, print usage and error
+if [[ -z "${max}" ]]; then
+  usage;
+fi
+
+# Don't specify the version with the go prefix, just 1.x.y will do.
+if ! printf '%s\n' "${max}" | grep -v -q "^go"; then
```
thockin, line 51: Simpler:

```
$ V="1.2.3"

$ W="go 1.2.3"

$ X=""

$ Y="something else"

$ Z="1.2.3.4"

$ for i in "$V" "$W" "$X" "$Y" "$Z"; do if [[ $i =~ ^[0-9]\.[0-9](\.[0-9])?$ ]]; then echo "'$i' matches"; else echo "'$i' does not match"; fi; done
'1.2.3' matches
'go 1.2.3' does not match
'' does not match
'something else' does not match
'1.2.3.4' does not match
```

So:

```
if [[ ! "${max}" =~ ^[0-9]\.[0-9](\.[0-9])?$ ]]; then
    usage
fi
```

hack/verify-go-directive.sh
```diff
@@ -0,0 +1,64 @@
...
+  usage;
+fi
+
+current=$(grep '^go [1-9]*' go.mod | cut -d ' ' -f2)
```
thockin, line 55: safer: `awk '/^go / {print $2;}' go.mod`

hack/verify-go-directive.sh
```diff
@@ -0,0 +1,64 @@
...
+if [[ -z "${current}" ]]; then
+  echo >&2 "FAIL: could not get value of go directive from go.mod"
+  exit 1
+fi
+
+if ! printf '%s\n' "${current}" "${max}" | sort --check=silent --version-sort; then
```
thockin, line 61: Why not test for exact equal?  We don't want it to go backwards either, right?

Makefile
```diff
@@ -4,6 +4,15 @@ all:
 
 test:
 	go test -race ./...
-	./hack/verify-examples.sh
 	go -C v2 test -race ./...
 	(cd v2 && ./hack/verify-examples.sh)
+
+verify: verify-go-directive
```
thockin, line 10: Do we need 2 levels of make rules?   Why not:

```
verify:
	./hack/verify-examples.sh
	./hack/verify-go-directive.sh 1.20
```

thockin (commented):
Will you also fix the presubmit job config to call `verify` ?

hack/verify-go-directive.sh
MadhavJivrajani: Trying to understand this a little better, would this serve as a sanity check in general? Because when we bump deps, the `go` directive is either going to stay the same or its going to increase, I don't think it can go backwards iiuc.

MadhavJivrajani: I understand now, thanks for laying it out @thockin!

thockin: Uggh, after making my point, you clever showed me I was being an idiot.  This is not "locking" the version, it really is upper-bounding it.  As soon as we add support for type-params, 1.13 will have to bump up, I guess?

thockin (commented):
Thanks!

/lgtm
/approve

hack/verify-go-directive.sh
MadhavJivrajani: Yeah, "upper-bounding" is a nice way to put it. Thanks! <3 

> As soon as we add support for type-params, 1.13 will have to bump up, I guess?

Yeah, that sounds about right!

Issue #266, opened 2024-03-10
TODO: Reconsider the "" package


thockin:
Gengo adds all pointer type and primitives, methods, etc to pkg "" - why?  Is this really useful to anyone?

PR #268, opened 2024-04-04
Handle empty package name from import tracker while naming a type.

mtrqq:
This commit adds handling for empty package namer while generating a type name, effectively treating such types as local ones.  Right now the behavior would generate an invalid syntax in such scenario

Probably related to https://github.com/kubernetes/gengo/issues/266

This PR is needed in the context of https://github.com/kubernetes/kubernetes/pull/124190 so that we are able to mitigate the problem with invalid package name supplied into namer without largely impacting other usage scenarios.

xref https://github.com/kubernetes/kubernetes/issues/124192

/assign jpbetz

### DIFF
--- a/v2/namer/namer_test.go
+++ b/v2/namer/namer_test.go
@@ -103,3 +103,32 @@ func TestNameStrategy(t *testing.T) {
 		t.Errorf("Wanted %#v, got %#v", e, a)
 	}
 }
+
+// NOTE: this test is intended to demostrate specific behavior
+// described in https://github.com/kubernetes/kubernetes/issues/124192
+// and should not be considered a good usage scenario of Namer
+//
+// Here we are trying to simulate the behavior when Namer is
+// configured with invalid local package name and import tracker
+// has valid configuration. In this case Namer shouldn't produce
+// invalid syntax like '.<variableName>' and instead treat empty
+// names from import tracker as types within the same package
+func TestRawNamerEmptyImportInteraction(t *testing.T) {
+	tracker := NewDefaultImportTracker(types.Name{
+		Package: "resource.io/pkg",
+		Name:    "name",
+	})
+	namer := NewRawNamer("resource/pkg", &tracker)
+
+	gotName := namer.Name(&types.Type{
+		Name: types.Name{
+			Package: "resource.io/pkg",
+			Name:    "flag",
+			Path:    "/path/resource.io/pkg",
+		},
+	})
+
+	if gotName != "flag" {
+		t.Errorf("Wanted %#v, got %#v", "flag", gotName)
+	}
+}
--- a/v2/namer/namer.go
+++ b/v2/namer/namer.go
@@ -329,7 +329,12 @@ func (r *rawNamer) Name(t *types.Type) string {
 			if t.Name.Package == r.pkg {
 				name = t.Name.Name
 			} else {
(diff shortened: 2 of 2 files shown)


jpbetz:
+@thockin

mtrqq:
> This makes sense to me given the example you provided out-of-band. @mtrqq Would you mind summarizing the example here?

Sure, added more context in the description

thockin:
> mitigate the problem with invalid package name supplied into namer

Can you help me understand if this is about "invalid" input (i.e. it should not work) or about valid input which is somehow different from anything we have seen in the past?

What would you think about writing a unit test for Namer?  That would help clarify what this is about...

jpbetz:
> > mitigate the problem with invalid package name supplied into namer
> 
> Can you help me understand if this is about "invalid" input (i.e. it should not work) or about valid input which is somehow different from anything we have seen in the past?
> 
> What would you think about writing a unit test for Namer? That would help clarify what this is about...

@mtrqq Could provide the planned use of this change? Showing how this ultimately is use to resolve https://github.com/kubernetes/kubernetes/issues/124192 might help here.

mtrqq:
@thockin @jpbetz Sure, added more context into a namer test. And a bit more story on this specific fix:

Currently we do have a bug within `applyconfig-gen` code generator because it configures Namer incorrectly and instead of `resource.io/pkg` supplies `resource/pkg`, we are not able to fix the direct cause of this issue as there are kubernetes packages which depend on this exact behavior with trimming package name with multiple tokens. Currently the plan is to fix only circular imports issue without touching other usecases, you may take a look inside https://github.com/kubernetes/kubernetes/pull/124190 which makes `ImportTracker` aware of the package it operates in, which leads us to `Namer` which currently produces invalid syntax in case of empty string result from `ImportTracker` (e.g. `.FooStruct` when we expect package name before dot)

This PR adds handling for empty package names from the `ImportTracker`, even in the worst case - it won't harm anyone as currently we still produce syntax errors in this case.

v2/namer/namer_test.go
```diff
@@ -103,3 +103,32 @@ func TestNameStrategy(t *testing.T) {
 		t.Errorf("Wanted %#v, got %#v", e, a)
 	}
 }
+
+// NOTE: this test is intended to demostrate specific behavior
+// described in https://github.com/kubernetes/kubernetes/issues/124192
+// and should not be considered a good usage scenario of Namer
```
jpbetz, line 109: This is making me wonder if we could fix the callsite(s) of `rawNamer.Name` instead of the function itself. Have we located the callsite(s)? Is there one or are there many?

thockin:
Is this the same as https://github.com/kubernetes/kubernetes/pull/124193 ?

jpbetz:
> Is this the same as https://github.com/kubernetes/kubernetes/pull/124193 ?

Yes, these are dupes.

thockin:
[kubernetes/kubernetes#124193](https://github.com/kubernetes/kubernetes/pull/124193) seems like the "right" fix, yes?

jpbetz:
> [kubernetes/kubernetes#124193](https://github.com/kubernetes/kubernetes/pull/124193) seems like the "right" fix, yes?

@mtrqq could you confirm it fixes your repro? I suspect it will.

PR #274, opened 2024-08-23
Fix deepcopy for generics

Jefftree:
We had a regression with the deepcopy gen not being able to handle generic fields.

```
type Foo struct {
   field set.Set[string]
}
```

This fixes the name to not be trimmed for generics in `goNameToName` and also fixes a bug with named interfaces capturing generics not capturing generics properly.

### DIFF
--- a/v2/go.mod
+++ b/v2/go.mod
@@ -10,5 +10,6 @@ require (
 
 require (
 	github.com/go-logr/logr v0.2.0 // indirect
+	github.com/google/go-cmp v0.6.0 // indirect
 	golang.org/x/mod v0.14.0 // indirect
 )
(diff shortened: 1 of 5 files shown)


thockin:
verified against k/k - this function seems like a great target for a unit test

Jefftree:
/assign @thockin 
/cc @TheSpiritXIII

Updated with fixes.

v2/go.mod
```diff
@@ -10,5 +10,6 @@ require (
 
 require (
 	github.com/go-logr/logr v0.2.0 // indirect
+	github.com/google/go-cmp v0.6.0 // indirect
```
Jefftree, line 13: This change is optional but I figured it'd be more useful to use `cmp.Diff` instead of a two complex structs that are quite hard to read in the tests. k/k already has this dependency https://github.com/kubernetes/kubernetes/blob/master/go.mod#L38 so we are not adding anything new for them.

thockin:
Thanks!

/lgtm
/approve

PR #276, opened 2024-08-23
Add in defaulter tests from prow

Jefftree:
Noticed that we specifically test defaulter-gen in prow and add it back in the Makefile. I believe it is not captured in `go test ./...` because the directory has a separate go mod file. 

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/gengo/gengo-config.yaml#L20

/assign @thockin @jpbetz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ test:
 	go test -race ./...
 	go -C v2 test -race ./...
 	(cd v2 && ./hack/verify-examples.sh)
+	(cd ./examples/defaulter-gen/_output_tests && go test -v -race ./...)
 
 # We verify for the maximum version of the go directive as 1.20
 # here because the oldest go directive that exists on our supported


Makefile
```diff
@@ -6,6 +6,7 @@ test:
 	go test -race ./...
 	go -C v2 test -race ./...
 	(cd v2 && ./hack/verify-examples.sh)
+	(cd ./examples/defaulter-gen/_output_tests && go test -v -race ./...)
```
thockin, line 9: why not run ./hack/verify-examples.sh in the root?

Jefftree: v2 has its own gomod which would change the go behavior whether this is run in the root vs in v2 directory right?

thockin: I don't understand what this means.

hack/verify-examples.sh removes generated code (in v1 only), then runs the generators (in v1 only), then runs tests (in v1 only).  That is inclusive of 

I am suggesting this rule should basically be:

```
test: testv1 testv2

testv1:
    go test -race ./...
    ./hack/verify-examples.sh

testv2:
    go -C v2 test -race ./...
    (cd v2 && ./hack/verify-examples.sh)

thockin: Oh, I see that's another go.mod.  Sigh.  Editing comment

thockin: Also, keep in mind v2 is alive, v1 is in stasis. :)

Jefftree:
I think this is covered by https://github.com/kubernetes/gengo/pull/279 and will close this

thockin:
Sorry!  I forgot this was still open.

PR #278, opened 2024-09-04
Ensure "make verify" verifies v2

skitt:


thockin (commented):
Why do we have "test" and "verify" and overlap between them?

I am going to OK this and offer a followup

thockin:
Thanks!

/lgtm
/approve

PR #280, opened 2024-09-05
Remove legacy gengo

skitt:
Legacy users have no particular reason to upgrade unless they switch to v2. This ensures we don't waste time on v1.

All the examples have already been moved to k/k.


jpbetz:
My understanding v2 is for kubernetes developers alone. It's not something we're planning to offer support guarantees for.  The generators we build with it we support, but gengo/v2 itself is for us.  Does that influence our decision here?

skitt:
> My understanding v2 is for kubernetes developers alone. It's not something we're planning to offer support guarantees for. The generators we build with it we support, but gengo/v2 itself is for us. Does that influence our decision here?

Yup, that’s why I moved the v2 README to the project root. The only reason I can see to keep v1 is if anyone would ever want to fix something in v1 for the benefit of external users of v1.

skitt:
In fact, it seems to me that gengo/v2 would benefit from being moved to an internal package in k/k code-generator...

thockin:
When I was doing the workspaces project, we did discuss "repatriating" gengo into k/k but decided it was more work than it was worth at the time.  We could move it to staging, so we keep the external repo in a similar state. Or we could just move it to an internal package, no module, which is simpler in some ways and more complicated in others.

Short of that though, I'm not sure how I feel about deleting V1.  It's not something I am eager to support, but we did put it out there and it costs us very little to leave it.  Having been the victim of "We're not supporting this anymore, so we deleted it", My inclination is just to leave it alone.  

Is there some other motivating factor that I'm not considering?

skitt:
> Is there some other motivating factor that I'm not considering?

Just general clean-up and clarifying the support status. What concerned me really was the existence of example code here with no corresponding examples in v2 (since those examples are now in k/k).

thockin:
Call me paranoid, but I'm more inclined to just leave it, at least for a while longer.  Can we revisit this in Q1'25?

PR #281, opened 2024-09-10
Preserve existing behavior with go1.22/1.23 alias type parsing

liggitt:
go/types returns different content for aliases in go1.23 by default, which breaks or unexpectedly modifies the type parsing / walking encountered by generators.

https://tip.golang.org/doc/go1.23#gotypespkggotypes

> By default, go/types now produces [Alias](https://tip.golang.org/pkg/go/types#Alias) type nodes for type aliases. This behavior can be controlled by the GODEBUG gotypesalias flag. Its default has changed from 0 in Go 1.22 to 1 in Go 1.23.

Encountered when bumping k/k to go 1.23 - https://github.com/kubernetes/kubernetes/pull/127271#issuecomment-2341698484


liggitt:
Testing in https://github.com/kubernetes/kubernetes/pull/127271

v2/parser/parse.go
```diff
@@ -486,7 +487,7 @@ func (p *Parser) addPkgToUniverse(pkg *packages.Package, u *types.Universe) erro
 	for _, n := range s.Names() {
 		switch obj := s.Lookup(n).(type) {
 		case *gotypes.TypeName:
-			t := p.walkType(*u, nil, obj.Type())
+			t := p.walkType(*u, nil, unalias(obj.Type()))
```
liggitt, line 490: default behavior prior to go1.23 resolved top-level aliases to their underlying types ... this preserves that

v2/parser/parse_122.go
```diff
@@ -0,0 +1,53 @@
...
+	case *gotypes.Alias:
+		// handle alias types specifically (only in go1.22, default in go1.23+)
+		name := goNameToName(t.String())
+		out := u.Type(name)
+		if out.Kind != types.Unknown {
+			return out
+		}
+
+		switch unaliased := gotypes.Unalias(t).(type) {
+		case *gotypes.Basic, *gotypes.Map, *gotypes.Slice:
```
liggitt, line 43: only these three types resulted in nodes of kind=Alias previously, other types were resolved directly to their underlying type

thockin (commented):
Looks correct.  We should just rename gengo's own "alias" to "typedef".

/lgtm
/approve

sayboras:
I came across this PR while working on cilium/cilium repo. Thanks a lot for your awesome work.

Not sure if this is the right place, I am just curious if k8s.io/gengo (not k8s.io/gengo/v2) is still supported. If yes, do we need to do the same for k8s.io/gengo package? Thanks.

thockin:
gengo is EOL and will not receive such updates.

gengo/V2 has significantly changed internally, and we opted to declare it a "virtually private" repo, primarily for use in kubernetes.  IOW, the compatibility contract is more or less waived.

Is cilium using gengo?  I was not aware of that (didn't look too hard, honestly).  If so, can you tell us more about the use-case?

sayboras:
Thanks for your reply.

> Is cilium using gengo? I was not aware of that (didn't look too hard, honestly). If so, can you tell us more about the use-case?

Cilium is using a forked [deepequal](https://github.com/cilium/deepequal-gen), which still relies on gengo v1. It seems like the cleanest path is to upgrade to gengo/v2.

thockin:
I see. I want to repeat, just so there's no confusion, gengo v2 is not really a supported api, and if you use it outside of the kubernetes project we're not making very strong API guarantees.  That doesn't mean we're going to break it on purpose, but it might mean that we are a little cavalier about changes.  

For example, we didn't really write a conversion guide for V1 users. So as you undertake this, if you undertake this, let me know if you struggle to figure out what changed.

mtardy:
Any chance we could get this fix into https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/code-generator? Controller-gen is now fixed to work with the new Alias type but not code-generator yet.

PR #285, opened 2024-12-20
Don't identity-alias stdlib imports

brandond:
Fixes stuff like `context "context"` and `http "net/http"`.

Ref:
* https://github.com/kubernetes/gengo/pull/277#issuecomment-2557434902


brandond:
@thockin any chance of getting this upstreamed or should I just close it?

thockin:
I'm not sure this is sufficient.  It's rare but sometimes the package name and the leaf of the path are not the same.  Looking at this code, I doubt we handle it properly in the absence of an alias like this.

This is one area where my big refactoring last year dared not go - the importer logic is twisty.  I figured it's generated code, so who cares, really.  I am not against fixing it, but I think that a change like this will explode in the above case.

@jpbetz

brandond:
Opened https://github.com/kubernetes/kubernetes/pull/132661 @jpbetz @thockin

thockin:
The goal is to get a CI run that passes k/k (or fails only for reasons explained by the vedor hackery).

Unfortunately, that PR is huge and already needs a rebase. :(

brandond:
am I running the correct commands to touch everything that would need touching?

thockin:
I could not shake the feeling that this was broken for cases where the dirname and pkgname did not match, so I hacked (seriously ugly) a test case.  Yeah, it's broken.

```diff
diff --git a/v2/examples/pointuh/main.go b/v2/examples/pointuh/main.go
index a469f38..5d5fe40 100644
--- a/v2/examples/pointuh/main.go
+++ b/v2/examples/pointuh/main.go
@@ -163,6 +163,7 @@ func getTargets(c *generator.Context, args *Args) []generator.Target {
 type pointuhGenerator struct {
 	generator.GoGenerator
 	myPackage *types.Package
+	imports   namer.ImportTracker
 }
 
 func newPointuhGenerator(outputFilename string, pkg *types.Package) generator.Generator {
@@ -171,6 +172,7 @@ func newPointuhGenerator(outputFilename string, pkg *types.Package) generator.Ge
 			OutputFilename: outputFilename,
 		},
 		myPackage: pkg,
+		imports:   generator.NewImportTrackerForPackage(pkg.Name),
 	}
 }
 
@@ -181,6 +183,7 @@ func (g *pointuhGenerator) Namers(*generator.Context) namer.NameSystems {
 	return namer.NameSystems{
 		// This elides package names when the name is in "this" package.
 		"localraw": namer.NewRawNamer(g.myPackage.Path, nil),
+		"raw":      namer.NewRawNamer("", g.imports),
 	}
 }
 
@@ -203,13 +206,33 @@ func (g *pointuhGenerator) GenerateType(c *generator.Context, t *types.Type, w i
 	return sw.Error()
 }
 
+func (g *pointuhGenerator) Imports(_ *generator.Context) (imports []string) {
+	var importLines []string
+	for _, singleImport := range g.imports.ImportLines() {
+		importLines = append(importLines, singleImport)
+	}
+	return importLines
+}
+
+var done = false
+
 func emitUnmodifiedType(t *types.Type, sw *generator.SnippetWriter) {
 	if t.Kind == types.DeclarationOf || t.Kind == types.Interface {
 		return
 	}
 
 	args := argsFromType(t)
+	args["foo"] = &types.Type{
+		Name: types.Name{
+			Name:    "Foo",
+			Package: "k8s.io/gengo/v2/foodir",
+		},
+	}
 	sw.Do("// $.type|localraw$ is an autogenerated clone of $.type|raw$\n", args)
+	if !done {
+		sw.Do("type foo $.foo|raw$ \n", args)
+		done = true
+	}
 	sw.Do("type $.type|localraw$ ", args)
 	for {
 		if t.Kind != types.Pointer {
diff --git a/v2/foodir/foo.go b/v2/foodir/foo.go
new file mode 100644
index 0000000..76c9182
--- /dev/null
+++ b/v2/foodir/foo.go
@@ -0,0 +1,3 @@
+package foopkg
+
+type Foo struct{}
diff --git a/v2/generator/import_tracker.go b/v2/generator/import_tracker.go
index 22393e4..1931636 100644
--- a/v2/generator/import_tracker.go
+++ b/v2/generator/import_tracker.go
@@ -47,7 +47,13 @@ func NewImportTrackerForPackage(local string, typesToAdd ...*types.Type) *namer.
 	tracker := namer.NewDefaultImportTracker(types.Name{Package: local})
 	tracker.IsInvalidType = func(*types.Type) bool { return false }
 	tracker.LocalName = func(name types.Name) string { return goTrackerLocalName(&tracker, local, name) }
-	tracker.PrintImport = func(path, name string) string { return name + " \"" + path + "\"" }
+	tracker.PrintImport = func(path, name string) string {
+		dirs := strings.Split(path, namer.GoSeparator)
+		if len(dirs) > 0 && name == dirs[len(dirs)-1] {
+			return "\"" + path + "\""
+		}
+		return name + " \"" + path + "\""
+	}
 
 	tracker.AddTypes(typesToAdd...)
 	return &tracker
@@ -61,13 +67,13 @@ func goTrackerLocalName(tracker namer.ImportTracker, localPkg string, t types.Na
 	path := t.Package
 
 	// Using backslashes in package names causes gengo to produce Go code which
-	// will not compile with the gc compiler. See the comment on GoSeperator.
+	// will not compile with the gc compiler. See the comment on GoSeparator.
 	if strings.ContainsRune(path, '\\') {
 		klog.Warningf("Warning: backslash used in import path '%v', this is unsupported.\n", path)
 	}
 	localLeaf := filepath.Base(localPkg)
 
-	dirs := strings.Split(path, namer.GoSeperator)
+	dirs := strings.Split(path, namer.GoSeparator)
 	for n := len(dirs) - 1; n >= 0; n-- {
 		// follow kube convention of not having anything between directory names
 		name := strings.Join(dirs[n:], "")
diff --git a/v2/namer/namer.go b/v2/namer/namer.go
index bae2ee9..2202f8e 100644
--- a/v2/namer/namer.go
+++ b/v2/namer/namer.go
@@ -26,14 +26,17 @@ import (
 )
 
 const (
-	// GoSeperator is used to split go import paths.
+	// GoSeparator is used to split go import paths.
 	// Forward slash is used instead of filepath.Seperator because it is the
 	// only universally-accepted path delimiter and the only delimiter not
 	// potentially forbidden by Go compilers. (In particular gc does not allow
 	// the use of backslashes in import paths.)
 	// See https://golang.org/ref/spec#Import_declarations.
 	// See also https://github.com/kubernetes/gengo/issues/83#issuecomment-367040772.
-	GoSeperator = "/"
+	GoSeparator = "/"
+	// GoSeperator is a typo for GoSeparator.
+	// Deprecated: use GoSeparator instead.
+	GoSeperator = GoSeparator
 )
 
 // Returns whether a name is a private Go name.
@@ -200,7 +203,7 @@ var (
 
 // filters out unwanted directory names and sanitizes remaining names.
 func (ns *NameStrategy) filterDirs(path string) []string {
-	allDirs := strings.Split(path, GoSeperator)
+	allDirs := strings.Split(path, GoSeparator)
 	dirs := make([]string, 0, len(allDirs))
 	for _, p := range allDirs {
 		if ns.IgnoreWords == nil || !ns.IgnoreWords[p] {
```

I fixed a typo in the process, sorry for the noise.

When I run this against the simple case (`go run ./examples/pointuh/ --output-dir ./footest --output-pkg example.com/footest ./examples/pointuh/testdata/simple/`), I get:

```
import (
        "k8s.io/gengo/v2/foodir"
)
```

and

```
type foo foodir.Foo
```

That's clearly not correct -- the package's name is "foopkg".  We either need the namer to emit `foopkg` or the importline to alias back to `foodir` (even though it seems redundant).  

We could maybe put a link to the `packages.Package` in each `Type`, so the import printer could see that the name and dir don't match, and alias it back to the dir.

Or we could teach namers how to look up the packages own name.  This is almost certainly more correct, but probably STILL ends up putting a link to the `packages.Package` in each `Type`.  There are so many layers of indirection in this area that it's hard to tell from reading what would be needed or what would be impacted if we changed it (and at what point in the lifecycle these things really happen).

Or we could just not.  Maybe the best fix here is just a comment documenting this as "load-bearing ugliness"?

thockin:
If you want to poke at it, you at least have a (hacktastic) test.  We should put a real test in for this, of course.

I would love to see an overhaul of Namers and Imports.  I didn't tackle it during the v1->v2 transition because it was already just SO MUCH change.  I don't know this part of the code that well, and I don't know what the resault should look like, so I don't even have a whole lot of guidance to offer.

brandond:
/remove-lifecycle rotten

Still working on this

brandond:
Still in my queue

PR #289, opened 2025-01-30
Add types.Any and types.Pointer

thockin:


jpbetz:
/approve
/lgtm
The use in validation-gen makes me feel confident about these

PR #290, opened 2025-01-30
Add Type.IsComparable()

thockin:


jpbetz:
/approve
/lgtm

I believe I fully grok this.  The `IsComparable` is needed in validation-gen to know if golang `comparable` may be used or if comparisions must be performed via a deep equals check.  So the `GoType` is introduced to track this and is ignored in test during `cmp.Equal()` since the equality should be based on the pre-existing fiels.

v2/parser/parse_test.go
```diff
@@ -1118,8 +1119,14 @@ func TestStructParse(t *testing.T) {
 			if st == nil || st.Kind == types.Unknown {
 				t.Fatalf("type %s not found", expected.Name.Name)
 			}
-			if e, a := expected, st; !reflect.DeepEqual(e, a) {
-				t.Errorf("wanted, got:\n%#v\n%#v\n%s", e, a, cmp.Diff(e, a))
+			if st.GoType == nil {
+				t.Errorf("type %s did not have GoType", expected.Name.Name)
+			}
+			opts := []cmp.Option{
+				cmpopts.IgnoreTypes(types.Type{}, "GoType"),
```
shashankram, line 1126: @thockin why is types.Type ignored in the comparison? This silently ignores the expected fields set in `expected`

thockin: There are non-public fields in `types.Type` which causes `cmp.Equal()` to panic.

That said, I think I meant to use `IgnoreFields` -- the arguments make sense for that and o not make sense for `IgnoreTypes`.

Do you want to send a PR, or shall I?

shashankram: https://github.com/kubernetes/gengo/pull/293

PR #293, opened 2025-05-13
parser/test: fix bug in comparison

shashankram:
Uses IgnoreFields instead of IgnoreTypes to ensure the expectations are correctly asserted. The assertion was broken by 1244d31, which caused all fields in `expected` to be ignored and not just `GoType`.


thockin (commented):
Thanks!

/lgtm
/approve

PR #294, opened 2025-05-13
v2/parser: prefer original comment when evaluating aliased Type

### LINKED ISSUE #292: Incorrect Descriptions are generated for an Aliased Type defined in multiple API versions
`gengo` produces flaking descriptions for the same type that is defined in multiple pkgs (using an alias), when used with openapi-gen (from kube-openapi).

In v1alpha2:
```
// BackendRef defines how a Route should forward a request to a Kubernetes
// resource.
//
// Note that when a namespace different than the local namespace is specified, a
// ReferenceGrant object is required in the referent namespace to allow that
// namespace's owner to accept the reference. See the ReferenceGrant
// documentation for details.
// +k8s:deepcopy-gen=false
type BackendRef = v1.BackendRef
```

In v1:
```
// BackendRef defines how a Route should forward a request to a Kubernetes
// resource.
//
// Note that when a namespace different than the local namespace is specified, a
// ReferenceGrant object is required in the referent namespace to allow that
// namespace's owner to accept the reference. See the ReferenceGrant
// documentation for details.
//
// <gateway:experimental:description>
//
// When the BackendRef points to a Kubernetes Service, implementations SHOULD
// honor the appProtocol field if it is set for the target Service Port.
//
// Implementations supporting appProtocol SHOULD recognize the Kubernetes
// Standard Application Protocols defined in KEP-3726.
//
// If a Service appProtocol isn't specified, an implementation MAY infer the
// backend protocol through its own means. Implementations MAY infer the
// protocol from the Route type referring to the backend Service.
//
// If a Route is not able to send traffic to the backend using the specified
// protocol then the backend is considered invalid. Implementations MUST set the
// "ResolvedRefs" condition to "False" with the "UnsupportedProtocol" reason.
//
// </gateway:experimental:description>
//
// Note that when the BackendTLSPolicy object is enabled by the implementation,
// there are some extra rules about validity to consider here. See the fields
// where this struct is used for more information about the exact behavior.
type BackendRef struct {
```

The following command generates different Descriptions for the same type in `zz_generated.openapi.go`:
```
go tool openapi-gen --output-file zz_generated.openapi.go --report-filename /tmp/generate.sh.api_violations.CUlQnn --output-dir /home/runner/work/kgateway/kgateway/pkg/generated/openapi --output-pkg github.com/kgateway-dev/kgateway/v2/pkg/generated/openapi github.com/kgateway-dev/kgateway/v2/api/v1alpha1 sigs.k8s.io/gateway-api/apis/v1 sigs.k8s.io/gateway-api/apis/v1alpha2 k8s.io/apimachinery/pkg/apis/meta/v1 k8s.io/api/core/v1 k8s.io/apimachinery/pkg/runtime k8s.io/apimachinery/pkg/util/intstr k8s.io/apimachinery/pkg/api/resource k8s.io/apimachinery/pkg/version
```

The diff:
```
diff --git a/pkg/generated/openapi/zz_generated.openapi.go b/pkg/generated/openapi/zz_generated.openapi.go
index 52786b7..8a7e159 100644
--- a/pkg/generated/openapi/zz_generated.openapi.go
+++ b/pkg/generated/openapi/zz_generated.openapi.go
@@ -20767,7 +20767,7 @@ func schema_sigsk8sio_gateway_api_apis_v1_BackendRef(ref common.ReferenceCallbac
 	return common.OpenAPIDefinition{
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
-				Description: "BackendRef defines how a Route should forward a request to a Kubernetes resource.\n\nNote that when a namespace different than the local namespace is specified, a ReferenceGrant object is required in the referent namespace to allow that namespace's owner to accept the reference. See the ReferenceGrant documentation for details.",
+				Description: "BackendRef defines how a Route should forward a request to a Kubernetes resource.\n\nNote that when a namespace different than the local namespace is specified, a ReferenceGrant object is required in the referent namespace to allow that namespace's owner to accept the reference. See the ReferenceGrant documentation for details.\n\n<gateway:experimental:description>\n\nWhen the BackendRef points to a Kubernetes Service, implementations SHOULD honor the appProtocol field if it is set for the target Service Port.\n\nImplementations supporting appProtocol SHOULD recognize the Kubernetes Standard Application Protocols defined in KEP-3726.\n\nIf a Service appProtocol isn't specified, an implementation MAY infer the backend protocol through its own means. Implementations MAY infer the protocol from the Route type referring to the backend Service.\n\nIf a Route is not able to send traffic to the backend using the specified protocol then the backend is considered invalid. Implementations MUST set the \"ResolvedRefs\" condition to \"False\" with the \"UnsupportedProtocol\" reason.\n\n</gateway:experimental:description>\n\nNote that when the BackendTLSPolicy object is enabled by the implementation, there are some extra rules about validity to consider here. See the fields where this struct is used for more information about the exact behavior.",
 				Type:        []string{"object"},
 				Properties: map[string]spec.Schema{
 					"group": {
@@ -23242,7 +23242,7 @@ func schema_sigsk8sio_gateway_api_apis_v1_ParentReference(ref common.ReferenceCa
 	return common.OpenAPIDefinition{
 		Schema: spec.Schema{
 			SchemaProps: spec.SchemaProps{
-				Description: "ParentReference identifies an API object (usually a Gateway) that can be considered a parent of this resource (usually a route). The only kind of parent resource with \"Core\" support is Gateway. This API may be extended in the future to support additional kinds of parent resources, such as HTTPRoute.\n\nNote that there are specific rules for ParentRefs which cross namespace boundaries. Cross-namespace references are only valid if they are explicitly allowed by something in the namespace they are referring to. For example: Gateway has the AllowedRoutes field, and ReferenceGrant provides a generic way to enable any other kind of cross-namespace reference.\n\nThe API object must be valid in the cluster; the Group and Kind must be registered in the cluster for this reference to be valid.",
+				Description: "ParentReference identifies an API object (usually a Gateway) that can be considered a parent of this resource (usually a route). There are two kinds of parent resources with \"Core\" support:\n\n* Gateway (Gateway conformance profile) * Service (Mesh conformance profile, ClusterIP Services only)\n\nThis API may be extended in the future to support additional kinds of parent resources.\n\nThe API object must be valid in the cluster; the Group and Kind must be registered in the cluster for this reference to be valid.",
 				Type:        []string{"object"},
 				Properties: map[string]spec.Schema{
 					"group": {
```

The description being generated switches between the descriptions defined in the v1 and v1alph2 versions of the API. I am not sure if this is a bug in openapi-gen or gengo.

shashankram:
Fixes the comment parsing on a Type so that it does not get overwritten by an aliased Type with a different comment.

Resolves #292


v2/parser/parse.go
```diff
@@ -488,7 +488,12 @@ func (p *Parser) addPkgToUniverse(pkg *packages.Package, u *types.Universe) erro
 		switch obj := s.Lookup(n).(type) {
 		case *gotypes.TypeName:
 			t := p.walkType(*u, nil, obj.Type())
-			p.addCommentsToType(obj, t)
+			// If the resolved type belongs to a different package, it implies
+			// that it is a type alias, in which case we should not update the
+			// comment lines on the original type with that of the alias
+			if t.Name.Package == pkgPath {
```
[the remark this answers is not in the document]
liggitt: (can you at least add to the unit test for an in-package alias to make sure that works properly)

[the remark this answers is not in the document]
shashankram: That's a good catch. Initially I was checking if the type is `*types.Alias`, but that breaks down when `GODEBUG=gotypesalias=0` or Aliases are not supported (Go versions <1.22) which our tests run into with the default go.mod version.

I'll try to address this case.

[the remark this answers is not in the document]
yongruilin: Maybe you can add the check of alias logic with the same function name into both `parse_122.go` and `parse_pre_122.go`

v2/parser/parse_test.go
```diff
@@ -1153,3 +1198,37 @@ func TestGoNameToName(t *testing.T) {
 		})
 	}
 }
+
+func TestCommentsWithAliasedType(t *testing.T) {
+	for i := 0; i < 50; i++ {
```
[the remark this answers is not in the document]
shashankram: Not sure what you are suggesting. Don't we need to process imports in `addPkgToUniverse`?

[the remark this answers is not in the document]
shashankram: We should not make assumptions on the import order.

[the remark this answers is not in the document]
yongruilin: Because the inconsistent is due to the `pkg.Imports`. 
Since your added test case would always parse v1 first then v2, the result of your current test case is consistent for each loop, so no need to run it 50 times.

But if you bring in another package like above `package a`, the order could be `a, v1, v2` or `a, v2, v1`.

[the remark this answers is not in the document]
shashankram: Ah, you were referring to the test. Makes sense, I can update the test.

v2/parser/parse.go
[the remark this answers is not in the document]
shashankram: Unfortunately that isn't enough. Pre go1.22 or when `GODEBUG=gotypesalias=0`, the alias type will resolve to the actual type, but the only way to not overwrite the actual type's comment is with something like:
```
if t.Name.Package == pkgPath && t.Name.Name == n {
        p.addCommentsToType(obj, t)
}
```
However, this breaks for Generic types like `type Blah[T any] struct {}` where `n == Blah` and `t.Name.Name == Blah[T]`.

So I don't think there is a way to prevent comment overwrites when Type Alias is not supported or disabled.

shashankram:
@yongruilin @liggitt this is ready for review. I updated the code to not overwrite the original type's comments from an aliased type when gotypesalias is enabled, and to produce deterministic outputs when aliases are not supported. I am not too concerned about the latter because gotypesalias is enabled by default starting 1.23, and we can provide correctness of comments there, and a deterministic output otherwise.

v2/parser/parse.go
```diff
@@ -510,11 +516,13 @@ func (p *Parser) addPkgToUniverse(pkg *packages.Package, u *types.Universe) erro
 
 	// Add all of this package's imports.
 	importedPkgs := []string{}
-	for _, imp := range pkg.Imports {
-		if err := p.addPkgToUniverse(imp, u); err != nil {
+	// Iterate imports in a predictable order
+	for _, imp := range slices.Sorted(maps.Keys(pkg.Imports)) {
+		pkg := pkg.Imports[imp]
```
liggitt, line 521: avoid shadowing pkg, keep `imp` referring to the same thing?

```
// Iterate imports in a predictable order
for _, key := range slices.Sorted(maps.Keys(pkg.Imports)) {
  img := pkg.Imports[key]
```

v2/parser/parse.go
```diff
@@ -488,7 +490,11 @@ func (p *Parser) addPkgToUniverse(pkg *packages.Package, u *types.Universe) erro
 		switch obj := s.Lookup(n).(type) {
 		case *gotypes.TypeName:
 			t := p.walkType(*u, nil, obj.Type())
-			p.addCommentsToType(obj, t)
+			// If this is a Type Alias, avoid updating the comments on the original
+			// Type t since walkType always returns the original type.
+			if !isTypeAlias(obj.Type()) {
+				p.addCommentsToType(obj, t)
```
liggitt, line 496: I'm trying to reason through what it means that alias types can no longer have comments attached… I agree we don't want them stomping the original type's comments but I'm not sure the implications of ignoring the alias' comments.

At the very least, before merging this, we should pull it into a speculative dependency update PR in kubernetes/kubernetes and double check if/how this impacts type generation there

v2/parser/parse.go
shashankram: Done

v2/parser/parse.go
shashankram: `walkType` always returns an unaliased type for an Alias type using `walkAliasType` which resolves the type using `gotypes.Unalias(t)`, so we didn't add comments on an aliased type before either. Instead, it would clobber the comments on the original type.

shashankram: @liggit I tested codegen in k/k https://github.com/kubernetes/kubernetes/compare/master...shashankram:kubernetes:gengo-test
- Tool updates: https://github.com/kubernetes/kubernetes/commit/3e0af33958e7506c0f1fd17889e5fc42823daaf3
- New codegen diff: https://github.com/kubernetes/kubernetes/commit/4510b30d1f29130df861b99ef675139d312ddab0

Confirmed that the openapi-gen version installed in _output prints the version built from my fork of kube-openapi which uses the gengo changes from this PR's branch:
```
./_output/local/go/bin/openapi-gen 
TEST BINARY
2025/05/14 16:34:39 Arguments validation error: --output-dir must be specified
```

liggitt: Hrmm... I guess there were these possible interactions before:
1. generating packages containing both the alias and the original, the comments on the original would be clobbered; the generation of the original would use the alias' comments (bad) and the generation of the alias would use the alias' comments (good)
2. generating a single package containing an alias, but not generating the package containing the original; the comments on the original type would be clobbered (bad, but non-impacting because the original wasn't being generated), but the generation of the alias would use the alias' comments (good)
3. generating multiple packages / aliases to an original type, but not generating the package containing the original; the comments on the original type would be clobbered by a non-deterministic alias (bad, but non-impacting because the original wasn't being generated), but the generation of all the aliases would use a non-deterministic alias' comments (good for that one, bad for the others)

So previous logic:
1. mixed good/bad (good for alias, bad for original)
2. good (single alias uses its own comments to generate)
3. mixed good/bad (good for one alias, bad for other aliases)

With this PR:
1. mixed good/bad (good for original, bad for alias)
2. bad (single alias has its comments ignored)
3. bad (all aliases have their comments ignored)

It's hard to say this is strictly better. Can we make all three scenarios better, isolating or shallow-copying the object resolved from aliases so that each alias can set its own comments without impacting the original?

shashankram: That's a good suggestion. I'll explore that

shashankram: @liggitt was your assessment based on the diff generated in https://github.com/kubernetes/kubernetes/commit/4510b30d1f29130df861b99ef675139d312ddab0?

I am not sure if your proposal is feasible if the Alias type is meant to resolve to the unaliased type but use different comments. The type names are stored in the Universe in a map:
```
type Universe map[string]*Package

type Package struct {
---
Types map[string]*Type
}
```

These are the scenarios:
1. gotypesalias=0 or not supported
`obj.Type()` will always resolve to the original type since a `(types.Scope).Lookup()` in Go will always resolve to the original type. So the alias is never tracked by its name separately in the Universe. This is true for Aliases regardless of whether they use the same or different name compared to the original type. 
2. gotypesalias=1
`obj.Type()` will resolve to `types.Alias` that is unaliased with `types.Unalias(t)` and walked.  In this case, we can prevent comment clobbering since we know that the type is an alias.

In both cases, the alias type does not exist in our Universe. We can start tracking aliases separately (not sure if this is possible when gotypesalias=0 as we would need to rely on pkg and type Name matching which is tricky for generics), but that would differ from our current generation. Is that desirable?

liggitt: Maybe the simplest thing to do is to require that aliases either 
1. Have no CommentLines / SecondClosestCommentLines, and preserve the comments from the original
2. Have identical CommentLines / SecondClosestCommentLines to the original, so it doesn't matter if we overwrite

something like 
```diff
--- a/v2/parser/parse.go
+++ b/v2/parser/parse.go
@@ -24,6 +24,7 @@ import (
 	"go/token"
 	gotypes "go/types"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
@@ -384,9 +385,30 @@ func (p *Parser) NewUniverse() (types.Universe, error) {
 
 // addCommentsToType takes any accumulated comment lines prior to obj and
 // attaches them to the type t.
-func (p *Parser) addCommentsToType(obj gotypes.Object, t *types.Type) {
-	t.CommentLines = p.docComment(obj.Pos())
-	t.SecondClosestCommentLines = p.priorDetachedComment(obj.Pos())
+func (p *Parser) addCommentsToType(obj gotypes.Object, t *types.Type) error {
+	if lines := p.docComment(obj.Pos()); len(lines) > 0 {
+		if len(t.CommentLines) > 0 && !reflect.DeepEqual(t.CommentLines, lines) {
+			return fmt.Errorf(
+				"cannot overwrite comments for type %v from:\n  %v\nto:\n  %v",
+				t.GoType,
+				strings.Join(t.CommentLines, "\n  "),
+				strings.Join(lines, "\n  "),
+			)
+		}
+		t.CommentLines = lines
+	}
+	if lines := p.priorDetachedComment(obj.Pos()); len(lines) > 0 {
+		if len(t.SecondClosestCommentLines) > 0 && !reflect.DeepEqual(t.SecondClosestCommentLines, lines) {
+			return fmt.Errorf(
+				"cannot overwrite secondClosestCommentLines for type %v from:\n  %v\nto:\n  %v",
+				t.GoType,
+				strings.Join(t.SecondClosestCommentLines, "\n  "),
+				strings.Join(lines, "\n  "),
+			)
+		}
+		t.SecondClosestCommentLines = lines
+	}
+	return nil
 }
 
 // packageDir tries to figure out the directory of the specified package.
@@ -488,21 +510,29 @@ func (p *Parser) addPkgToUniverse(pkg *packages.Package, u *types.Universe) erro
 		switch obj := s.Lookup(n).(type) {
 		case *gotypes.TypeName:
 			t := p.walkType(*u, nil, obj.Type())
-			p.addCommentsToType(obj, t)
+			if err := p.addCommentsToType(obj, t); err != nil {
+				return err
+			}
 		case *gotypes.Func:
 			// We only care about functions, not concrete/abstract methods.
 			if obj.Type() != nil && obj.Type().(*gotypes.Signature).Recv() == nil {
 				t := p.addFunction(*u, nil, obj)
-				p.addCommentsToType(obj, t)
+				if err := p.addCommentsToType(obj, t); err != nil {
+					return err
+				}
 			}
 		case *gotypes.Var:
 			if !obj.IsField() {
 				t := p.addVariable(*u, nil, obj)
-				p.addCommentsToType(obj, t)
+				if err := p.addCommentsToType(obj, t); err != nil {
+					return err
+				}
 			}
 		case *gotypes.Const:
 			t := p.addConstant(*u, nil, obj)
-			p.addCommentsToType(obj, t)
+			if err := p.addCommentsToType(obj, t); err != nil {
+				return err
+			}
 		default:
 			klog.Infof("addPkgToUniverse %q: unhandled object of type %T: %v", pkgPath, obj, obj)
 		}
```

liggitt: I mostly want to avoid silently overwriting the original *or* ignoring divergent comments on the aliases

shashankram: This would be a breaking change so we would need to introduce an option to opt-in to this behavior. Is there a downside to ignoring comments on the alias because the alias type does not exist in the Universe anyway?

liggitt: ignoring comments on the alias is also a breaking change, depending on what comments they had in place

shashankram: It wouldn't break code generation, it would simply produce a comment from the original type. What you are proposing will cause codegen to break and make API authors update the comments.

shashankram: Ah because of the codegen markers I guess. I am guessing this is a problem even today, as different markers on an Alias would overwrite the markers on the original type via comments?

My main concern with enforcing same comments on Alias and Original types is that consumers of APIs will be tied down to an older version of gengo till all APIs they consume meet the same/no comment requirement for aliased types, and they are not going to be able to control external APIs outside their project.

So we probably already have such instances where Aliased comments are negatively affecting codegen for the original Type silently, and the only viable fix is to ensure same/no comment on the alias.

liggitt: > Ah because of the codegen markers I guess.

Right.

> I am guessing this is a problem even today, as different markers on an Alias would overwrite the markers on the original type via comments?

If the generation is also happening for the original types, yes.

> My main concern with enforcing same comments on Alias and Original types is that consumers of APIs will be tied down to an older version of gengo till all APIs they consume meet the same/no comment requirement for aliased types, and they are not going to be able to control external APIs outside their project.

Hmm... any package gengo is going to write *to* should be under the user's control to fix, right? It looks like we only get to the spot where we call addCommentsToType for types in "userRequested" packages ... are those not in their control to fix source of?

shashankram: > Hmm... any package gengo is going to write to should be under the user's control to fix, right? It looks like we only get to the spot where we call addCommentsToType for types in "userRequested" packages ... are those not in their control to fix source of?

Not necessarily. E.g. https://github.com/kgateway-dev/kgateway/blob/main/hack/generate.sh#L53 where we call openapi-gen on `sigs.k8s.io/gateway-api/apis/*` which is an external project. In fact, this external project has a type alias with a different comment that caused codegen to flake. 

The `zz_generated.openapi.go` is consumed by `applyconfiguration-gen` via https://github.com/kgateway-dev/kgateway/blob/main/cmd/modelschema/main.go#L11.

Similarly, gateway-api calls openapi-gen on k8s types:
https://github.com/kubernetes-sigs/gateway-api/blob/6fbbd90954c2a30a6502cbb22ec6e7f3359ed3a0/hack/update-clientset.sh#L64

liggitt: Is there any kind of signal in the gengo invocation about which packages are considered *output* packages that we could use to decide whether to just warn or error when we detect an attempt to overwrite type comments?

liggitt: > `type v1.Foo = v2.Foo` or `type v1.FooAlias = v2.Foo`, generation will never produce the aliased type in the v1 pkg.

good point ... in fact, a cross-package type alias *cannot* declare new methods

Ok ... I think we *do* want to detect and warn if an alias is setting comments that differ from its underlying type but not error (and when we pull this into kubernetes/kubernetes resolve those warnings by removing or reconciling comments).

liggitt: something like https://github.com/kubernetes/gengo/pull/294/files#discussion_r2091473215 is probably still what I would try to do, but see if we can warn instead of erroring

shashankram: If the sorted import processing resolves the flaking codegen, then a warning may be acceptable to begin with. Though I think it would be a breaking change anyway as the sorted import processing during generation could change the generated openapi comment descriptions and validation constraints? If codegen still flakes, then a warning wouldn't suffice.

liggitt: if we warn _and avoid overwriting_, that should prevent flakes, right?

shashankram: Avoiding overwriting might not be preventable just from your suggestion in https://github.com/kubernetes/gengo/pull/294/files#discussion_r2091473215. To truly avoid overwriting, we need to reliably detect that a type is an Alias and it should not overwrite the original type. I am not sure if we will always evaluate an alias after the original type has been parsed. How do you propose we do that?

liggitt: When we see a mismatch:
* if the current object is an alias, warn and ignore
* if the current object is not an alias, warn and stomp

Something like this:

```diff
```diff --git a/v2/parser/parse.go b/v2/parser/parse.go
index 4c1efa00104..1f02613291a 100644
--- a/v2/parser/parse.go
+++ b/v2/parser/parse.go
@@ -24,6 +24,7 @@ import (
 	"go/token"
 	gotypes "go/types"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
@@ -385,8 +386,61 @@ func (p *Parser) NewUniverse() (types.Universe, error) {
 // addCommentsToType takes any accumulated comment lines prior to obj and
 // attaches them to the type t.
 func (p *Parser) addCommentsToType(obj gotypes.Object, t *types.Type) {
-	t.CommentLines = p.docComment(obj.Pos())
-	t.SecondClosestCommentLines = p.priorDetachedComment(obj.Pos())
+	if newLines, oldLines := p.docComment(obj.Pos()), t.CommentLines; len(newLines) > 0 {
+		switch {
+		case len(oldLines) == 0, reflect.DeepEqual(oldLines, newLines):
+			// no comments associated, or comments match exactly
+			t.CommentLines = newLines
+
+		case isTypeAlias(obj.Type()):
+			// ignore mismatched comments from obj because it's an alias
+			klog.Warningf(
+				"Mismatched comments seen for type %v.\n  Using comments:\n    %v\n  Ignoring comments from type alias %v:\n    %v",
+				t.GoType,
+				strings.Join(oldLines, "\n    "),
+				obj.Type(),
+				strings.Join(newLines, "\n    "),
+			)
+
+		case !isTypeAlias(obj.Type()):
+			// overwrite existing comments with ones from obj because obj is not an alias
+			t.CommentLines = newLines
+			klog.Warningf(
+				"Mismatched comments seen for type %v.\n  Using comments:\n    %v\n  Ignoring comments from type aliases:\n    %v",
+				t.GoType,
+				strings.Join(newLines, "\n    "),
+				strings.Join(oldLines, "\n    "),
+			)
+		}
+	}
+
+	if newLines, oldLines := p.priorDetachedComment(obj.Pos()), t.SecondClosestCommentLines; len(newLines) > 0 {
+		switch {
+		case len(oldLines) == 0, reflect.DeepEqual(oldLines, newLines):
+			// no comments associated, or comments match exactly
+			t.SecondClosestCommentLines = newLines
+
+		case isTypeAlias(obj.Type()):
+			// ignore mismatched comments from obj because it's an alias
+			klog.Warningf(
+				"Mismatched secondClosestCommentLines seen for type %v.\n  Using comments:\n    %v\n  Ignoring comments from type alias %v:\n    %v",
+				t.GoType,
+				strings.Join(oldLines, "\n    "),
+				obj.Type(),
+				strings.Join(newLines, "\n    "),
+			)
+
+		case !isTypeAlias(obj.Type()):
+			// overwrite existing comments with ones from obj because obj is not an alias
+			t.SecondClosestCommentLines = newLines
+			klog.Warningf(
+				"Mismatched secondClosestCommentLines seen for type %v.\n  Using comments:\n    %v\n  Ignoring comments from type aliases:\n    %v",
+				t.GoType,
+				strings.Join(newLines, "\n    "),
+				strings.Join(oldLines, "\n    "),
+			)
+		}
+	}
 }
 
 // packageDir tries to figure out the directory of the specified package.
@@ -557,7 +611,11 @@ func (p *Parser) priorCommentLines(pos token.Pos, lines int) *ast.CommentGroup {
 }
 
 func splitLines(str string) []string {
-	return strings.Split(strings.TrimRight(str, "\n"), "\n")
+	lines := strings.Split(strings.TrimRight(str, "\n"), "\n")
+	if len(lines) == 1 && lines[0] == "" {
+		return nil
+	}
+	return lines
 }
 
 func goFuncNameToName(in string) types.Name {
diff --git a/v2/parser/parse_122.go b/v2/parser/parse_122.go
index ec2064958a9..4b2c458c4c6 100644
--- a/v2/parser/parse_122.go
+++ b/v2/parser/parse_122.go
@@ -31,3 +31,8 @@ func (p *Parser) walkAliasType(u types.Universe, in gotypes.Type) *types.Type {
 	}
 	return nil
 }
+
+func isTypeAlias(in gotypes.Type) bool {
+	_, isAlias := in.(*gotypes.Alias)
+	return isAlias
+}
\ No newline at end of file
diff --git a/v2/parser/parse_pre_122.go b/v2/parser/parse_pre_122.go
index 6f62100c0a7..0249703b545 100644
--- a/v2/parser/parse_pre_122.go
+++ b/v2/parser/parse_pre_122.go
@@ -28,3 +28,7 @@ import (
 func (p *Parser) walkAliasType(u types.Universe, in gotypes.Type) *types.Type {
 	return nil
 }
+
+func isTypeAlias(in gotypes.Type) bool {
+	return false
+}
\ No newline at end of file
```

shashankram: Thanks for clarifying, I misunderstood your earlier comment as a suggestion to ignore alias detection.

> if the current object is an alias, warn and ignore
This would alter the generated schema and markers where it picked the Alias before. Is that acceptable now?

liggitt: I think we're saying it has to be acceptable, and if we want to avoid this behavior when we ingest this update into kubernetes/kubernetes, we would avoid it by modifying the original type's comments to match the alias comments that were getting honored before

liggitt: don't we want to still call `addCommentsToType` on aliases, and have it warn if the alias comments mismatch the original?

shashankram: Forgot to remove that check. Pushed the change to always call addCommentsToType

shashankram:
> Thanks, the PR looks good. Can you open the proof of concept PR against kubernetes/kubernetes that pulls this branch in as a temporary override of the gengo dependency, so we can double-check what alias comments we need to change to get CI green and code generation warning-free there before we merge here?

@liggitt here's the codegen diff: https://github.com/shashankram/kubernetes/commit/e2f5adf830344f26a6cf0fd7573f4dd00d023c8a

These are the types that have mismatched comments
```
Mismatched comments seen for type k8s.io/api/admissionregistration/v1.ReinvocationPolicyType. Using comments:
Mismatched comments seen for type k8s.io/api/admissionregistration/v1.ScopeType. Using comments:
Mismatched comments seen for type k8s.io/api/core/v1.IPFamilyPolicy. Using comments:
Mismatched comments seen for type k8s.io/api/core/v1.ServiceExternalTrafficPolicy. Using comments:
Mismatched comments seen for type k8s.io/api/core/v1.ServiceInternalTrafficPolicy. Using comments:
Mismatched comments seen for type k8s.io/apimachinery/pkg/apis/meta/v1.TableOptions. Using comments:
```

liggitt:
> [shashankram/kubernetes@b668816](https://github.com/shashankram/kubernetes/commit/b668816f7a7e9fd6ada96b69fe649939852184ca)

Thanks

/lgtm
/approve

> To clarify my understanding again, we are solving 2 issues in this change:
> 
> 1. With deterministic sorting of imports in `addPkgToUniverse`, we should get a deterministic output
> 2. Not clobber the comments(and markers) on the original type with that of an alias when gotypesalias=1, and warn on mismatch

Yes, exactly

shashankram:
@liggitt do we want more changes? If not, I'd like you to make another pass on this. I am also seeking an answer on https://github.com/kubernetes/gengo/pull/294#discussion_r2214305250

v2/parser/parse.go
liggitt: I don't think so, but if we find one, we can consider adjustments here

liggitt:
this lgtm, I don't have approve on this repo

thockin:
Oof, this is ugly for anything relying on comment tags.  Declarative validation...

This will work:

```
// +k8s:maxLength=63
// +k8s:format=ks8-short-name
type TheType string

// +k8s:maxLength=32
type TheAlias TheType
```

But this will not:

```
// +k8s:something1=63
// +k8s:something2
type TheType struct {
    S string
}

// +k8s:something1=32
type TheAlias TheType
```

I think?

thockin:
@jpbetz 

This is because gengo "helpfully" de-aliases structs and Go "helpfully" elides intermediate types.

This PR causers a TON of useless warnings in k/k.

thockin:
I seem to be wrong, so I must misunderstand what this is doing.

liggitt:
The warnings help see when mismatched comments are being ignored. I expected to resolve those by adjusting the comments on the aliases to match the original type comments.

thockin:
Some/many of those are like "Preserved for backwards compat" -- that seems useful to retain, but means they will forever be different.

thockin:
What if we allowed a delta of lines that start with "Deprecated:" so we can flag the "for-compat" ones and not see warnings?

PR #296, opened 2025-05-21
feat: add raw string support to gengo comment tag arg parsing

aaron-prindle:
Requirement for: https://github.com/jpbetz/validation-gen/pull/56

This PR adds raw string support to the gengo comment tag arg parser.  This allows for gengo to parse tags of the form:
```go
// +k8s:tagName(\`self.foo==bar, {"hello": "there"}, etc.\`)=...
```

  The args themselves keep the backtick character.  This allows for reducing ambiguity between raw string and identifier args:
```go
// +k8s:tagName(\`foo\`) -> arg=\`foo\` (not arg=foo)
```

This PR currently uses two scanners to accomplish this, one for splitting the gengo comment on the "top-level" "=":
```go
// +k8s:tagName(\`self.foo==bar, {"hello": "there"}, etc.\`)=...     // <- the right-most equals in this case
```
and another to parse the arg text:

\`self.foo==bar, {"hello": "there"}, etc.\`)

This PR minorly ratchets the args supported as it currently relies on scanner.Ident which enforces the first character is not a digit whereas the prevoius parser allows for the first character to be alphanumeric.  Original related code snippet below.  I believe this change is beneficial, does not affect usage in practice, and all original tests still pass.  The original parseTagArg docs mention Go-style identfier token so this should be aligned with using scanner.Ident:

https://github.com/kubernetes/gengo/blob/master/v2/comments.go#L271-L274
```go
	for i, r := range runes {
		if unicode.IsLetter(r) || unicode.IsDigit(r) {
			continue
		}
```

NOTE:  Currently this only adds raw string support but with the changes in this PR it would be simple to also add general string support (~3 line change not including tests).


v2/comments.go
```diff
@@ -260,31 +262,99 @@ func parseTagKey(input string, tagNames []string) (string, []string, error) {
...
+	// Expect no whitespace, etc. after the one ')'.
+	if s.Scan() != scanner.EOF {
+		pos := s.Pos().Offset - len(s.TokenText())
+		return nil, fmt.Errorf("unexpected characters after ')': %q", input[pos:])
+	}
+	return args, nil
+}
+
+type argScanner struct {
+	*scanner.Scanner
```
thockin, line 292: this is not an API I know and the docs are...not awesome.  Is it scanning Go's grammar or are we feeding it a BNF or something?

aaron-prindle: text/scanner acts as the lexer for a custom "grammar" the go code defines.  From the code in this PR, this is a BNF similar to:

TagArg ::= '(' [ SingleArgument ] ')'
SingleArgument ::= GoIdentifier | GoRawStringLiteral

eg: 
+tagArg(SingleArgument)=...

This is mainly defined with the below code snippets:
```
func initArgScanner(input string) *argScanner {
...
	s.Mode = scanner.ScanIdents | scanner.ScanRawStrings
```

```
	case scanner.RawString:
		return s.TokenText(), nil
	case scanner.Ident:
		txt := s.TokenText()
		for _, r := range txt {
			if !unicode.IsLetter(r) && !unicode.IsDigit(r) {
				return "", s.unexpectedTokenError("letter or digit", txt)
			}
		}
		return txt, nil
```

GoIdentifier has an additional limitation of being letters and digits only where Golang can use underscores in names as well (carried over from prior code).

NOTE: the code is modified now with the merge of https://github.com/kubernetes/gengo/pull/297

PR #297, opened 2025-05-21
Add JSON support to tag args for ExtractFunctionStyleCommentTags

jpbetz:


jpbetz:
@aaron-prindle This turned out to be easier to implement than I expected. Would you review?

Goal is not to support 100% of JSON.  The goal is to ensure that we only match a strict subset of json (hence the use of json unmarshal)

v2/comments_test.go
```diff
@@ -309,106 +319,143 @@ func TestParseTagKeyWithTagNames(t *testing.T) {
...
+			err:   true,
+		},
+		{
+			input: `"a`,
+			err:   true,
+		},
+		{
+			input: `UNKNOWN`,
+			err:   true,
+		},
```
aaron-prindle, line 439: I am seeing some potential small issues with the parser around whitespace.  Using the below test cases:

		{
			// name:  "whitespace before }",
			input: `{"key":"value" }`,
		},
		{
			// name:  "whitespace before ]",
			input: `[1 ]`,
		},
		{
			// name:  "whitespace before comma in array",
			input: `[1 ,2]`,
		},
I get errors for these cases when I believe they are valid JSON.  Might make sense to document this in the supported JSON subset or get these tests passing.  I am fine with either option.

v2/comments.go
```diff
@@ -229,132 +241,206 @@ func (t Tag) String() string {
...
 	}
-	// Expect no whitespace, etc. after the one ')'.
-	if s.Scan() != scanner.EOF {
-		pos := s.Pos().Offset - len(s.TokenText())
-		return nil, fmt.Errorf("unexpected characters after ')': %q", input[pos:])
+}
+
+// scanJSONFlavoredValue consumes a single token as a JSON value from the scanner and returns the token text.
+// A strict subset of JSON is supported, in particular:
+// - Big numbers and numbers with exponents are not supported.
```
aaron-prindle, line 292: nit: Can we add tests for these cases so that the subset is confirmed and codified in tests?

v2/comments_test.go
jpbetz: Fixed

v2/comments.go
jpbetz: Added

jpbetz:
Thanks @aaron-prindle. Feedback applied

PR #299, opened 2025-05-23
Add named arg parser

jpbetz:
Lots going on here:

- Add new ExtractFunctionStyleCommentTypedTags that supports both typed args and named args.(thanks @thockin for the parser!)
- All Extract functions call down to ExtractFunctionStyleCommentTypedTags to keep code paths simple.  But only the new function supports named params.
- Expand test coverage.
- Remove bare JSON support.

### DIFF
--- a/v2/argsparser.go
+++ b/v2/argsparser.go
@@ -19,11 +19,18 @@ const (
 	stEscape          = "stEscape"
 	stEndOfToken      = "stEndOfToken"
 	stMaybeComment    = "stMaybeComment"
+	stTrailingValue   = "stTrailingValue"
 	stTrailingSlash   = "stTrailingSlash"
 	stTrailingComment = "stTrailingComment"
 )
 
-func parseTagKey(input string) (string, []Arg, error) {
+type tagKey struct {
+	name       string
+	args       []Arg
+	valueStart int
+}
+
+func parseTagKey(input string) (tagKey, error) {
 	tag := bytes.Buffer{} // current tag name
 	args := []Arg{}       // all tag arguments
 
@@ -36,6 +43,8 @@ func parseTagKey(input string) (string, []Arg, error) {
 	var incomplete bool
 	var quote rune
 
+	valueStart := -1
+
 	isIdentBegin := func(r rune) bool {
 		return unicode.IsLetter(r) || r == '_'
 	}
@@ -47,7 +56,7 @@ func parseTagKey(input string) (string, []Arg, error) {
 		if _, err := strconv.ParseInt(s, 0, 64); err != nil {
 			return fmt.Errorf("invalid number %q", s)
 		}
-		cur.Typ = TypeInt
+		cur.Type = TypeInt
 		cur.Value = s
 		args = append(args, cur)
 		cur = Arg{}
@@ -56,7 +65,7 @@ func parseTagKey(input string) (string, []Arg, error) {
 	}
 	saveString := func() {
 		s := buf.String()
-		cur.Typ = TypeString
+		cur.Type = TypeString
 		cur.Value = s
 		args = append(args, cur)
 		cur = Arg{}
@@ -65,9 +74,9 @@ func parseTagKey(input string) (string, []Arg, error) {
 	saveBoolOrString := func() {
 		s := buf.String()
 		if s == "true" || s == "false" {
-			cur.Typ = TypeBool
+			cur.Type = TypeBool
 		} else {
-			cur.Typ = TypeString
+			cur.Type = TypeString
 		}
 		cur.Value = s
 		args = append(args, cur)
@@ -99,9 +108,13 @@ parseLoop:
 			switch {
 			case isIdentInterior(r):
 				tag.WriteRune(r)
+			case r == ':': // allowed in tag names
+				tag.WriteRune(r)
 			case r == '(':
 				incomplete = true
 				st = stArg
+			case r == '=':
+				st = stTrailingValue
 			case unicode.IsSpace(r):
 				st = stMaybeComment
 			default:
@@ -117,6 +130,9 @@ parseLoop:
 			case r == '0':
 				buf.WriteRune(r)
 				st = stPrefixedNumber
+			case r == '-':
+				buf.WriteRune(r)
+				st = stNumber
 			case unicode.IsDigit(r):
 				buf.WriteRune(r)
 				st = stNumber
@@ -137,18 +153,18 @@ parseLoop:
 				continue
 			case r == ',':
 				if err := saveInt(); err != nil {
-					return "", nil, err
+					return tagKey{}, err
 				}
 				st = stArg
 			case r == ')':
 				if err := saveInt(); err != nil {
-					return "", nil, err
+					return tagKey{}, err
 				}
 				incomplete = false
 				st = stMaybeComment
 			case unicode.IsSpace(r):
 				if err := saveInt(); err != nil {
-					return "", nil, err
+					return tagKey{}, err
 				}
 				st = stEndOfToken
 			default:
@@ -181,7 +197,7 @@ parseLoop:
 				buf.WriteRune(r)
 				st = stQuotedString
 			default:
-				return "", nil, fmt.Errorf("unhandled escaped character %q", r)
+				return tagKey{}, fmt.Errorf("unhandled escaped character %q", r)
 			}
 		case stNakedString:
 			switch {
@@ -222,9 +238,15 @@ parseLoop:
 			case r == '/':
 				incomplete = true
 				st = stTrailingSlash
+			case r == '=':
+				st = stTrailingValue
 			default:
 				break parseLoop
 			}
+		case stTrailingValue:
+			valueStart = i
+			i = len(runes) - 1
+			break parseLoop
 		case stTrailingSlash:
 			switch {
 			case r == '/':
@@ -241,22 +263,22 @@ parseLoop:
 		}
 	}
 	if i != len(runes)-1 {
-		return "", nil, fmt.Errorf("unexpected character %q at position %d", r, i)
+		return tagKey{}, fmt.Errorf("unexpected character %q at position %d", r, i)
 	}
 	if incomplete {
-		return "", nil, fmt.Errorf("unexpected end of input")
+		return tagKey{}, fmt.Errorf("unexpected end of input")
 	}
 	usingNamedArgs := false
 	for i, arg := range args {
 		if (usingNamedArgs && arg.Name == "") || (!usingNamedArgs && arg.Name != "" && i > 0) {
-			return "", nil, fmt.Errorf("can't mix named and positional arguments")
+			return tagKey{}, fmt.Errorf("can't mix named and positional arguments")
 		}
 		if arg.Name != "" {
 			usingNamedArgs = true
 		}
 	}
 	if !usingNamedArgs && len(args) > 1 {
-		return "", nil, fmt.Errorf("multiple arguments must use 'name: value' syntax")
+		return tagKey{}, fmt.Errorf("multiple arguments must use 'name: value' syntax")
 	}
-	return tag.String(), args, nil
+	return tagKey{name: tag.String(), args: args, valueStart: valueStart}, nil
 }
--- a/v2/argsparser.go
+++ b/v2/argsparser.go
@@ -1,0 +1,262 @@
+package gengo
+
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+	"strings"
+	"unicode"
+)
+
+const (
+	stBegin           = "stBegin"
+	stTag             = "stTag"
+	stArg             = "stArg"
+	stNumber          = "stNumber"
+	stPrefixedNumber  = "stPrefixedNumber"
+	stQuotedString    = "stQuotedString"
+	stNakedString     = "stNakedString"
+	stEscape          = "stEscape"
+	stEndOfToken      = "stEndOfToken"
+	stMaybeComment    = "stMaybeComment"
+	stTrailingSlash   = "stTrailingSlash"
+	stTrailingComment = "stTrailingComment"
+)
+
+func parseTagKey(input string) (string, []Arg, error) {
+	tag := bytes.Buffer{} // current tag name
+	args := []Arg{}       // all tag arguments
+
+	cur := Arg{}          // current argument accumulator
+	buf := bytes.Buffer{} // string accumulator
+
+	// These are defined outside the loop to make errors easier.
+	var i int
+	var r rune
+	var incomplete bool
+	var quote rune
+
+	isIdentBegin := func(r rune) bool {
+		return unicode.IsLetter(r) || r == '_'
+	}
+	isIdentInterior := func(r rune) bool {
+		return unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_'
+	}
+	saveInt := func() error {
+		s := buf.String()
+		if _, err := strconv.ParseInt(s, 0, 64); err != nil {
+			return fmt.Errorf("invalid number %q", s)
+		}
+		cur.Typ = TypeInt
+		cur.Value = s
+		args = append(args, cur)
+		cur = Arg{}
+		buf.Reset()
+		return nil
+	}
+	saveString := func() {
+		s := buf.String()
+		cur.Typ = TypeString
+		cur.Value = s
+		args = append(args, cur)
+		cur = Arg{}
+		buf.Reset()
+	}
+	saveBoolOrString := func() {
+		s := buf.String()
+		if s == "true" || s == "false" {
+			cur.Typ = TypeBool
+		} else {
+			cur.Typ = TypeString
+		}
+		cur.Value = s
+		args = append(args, cur)
+		cur = Arg{}
+		buf.Reset()
+	}
+	saveName := func() {
+		cur.Name = buf.String()
+		buf.Reset()
+	}
+
+	runes := []rune(input)
+	st := stBegin
+parseLoop:
+	for i, r = range runes {
+		//fmt.Printf("state: %s, char: %q\n", st, r)
+		switch st {
+		case stBegin:
+			switch {
+			case unicode.IsSpace(r):
+				continue
+			case isIdentBegin(r):
+				tag.WriteRune(r)
+				st = stTag
+			default:
+				break parseLoop
+			}
+		case stTag:
+			switch {
+			case isIdentInterior(r):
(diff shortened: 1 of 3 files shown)


jpbetz:
/assign @thockin 
This makes a few small additions to the parser.  Namely support for tag keys that end in `=` followed by a value and support for the `:` char in tag names.

v2/argsparser.go
```diff
@@ -99,9 +108,13 @@ parseLoop:
 			switch {
 			case isIdentInterior(r):
 				tag.WriteRune(r)
+			case r == ':': // allowed in tag names
+				tag.WriteRune(r)
```
jpbetz, line 112: @thockin We could expand the grammar from `<tagBody> ::= <name> ...` to `<tagBody> ::= { <prefix> ":" } <name> ...`.  But what I did here was to just allow ":" in tag names.

Thoughts welcome.

v2/argsparser.go
```diff
@@ -117,6 +130,9 @@ parseLoop:
 			case r == '0':
 				buf.WriteRune(r)
 				st = stPrefixedNumber
+			case r == '-':
+				buf.WriteRune(r)
+				st = stNumber
```
jpbetz, line 135: @thockin Should we allow negative numbers?  I've added it here.

v2/argsparser.go
```diff
@@ -222,9 +238,15 @@ parseLoop:
...
 				st = stTrailingSlash
+			case r == '=':
+				st = stTrailingValue
 			default:
 				break parseLoop
 			}
+		case stTrailingValue:
+			valueStart = i
+			i = len(runes) - 1
+			break parseLoop
```
jpbetz, line 249: @thockin The idea here was to track where the "beginning of the key value" is found so I can return the position of it in the parse result.

v2/comments.go
```diff
@@ -188,6 +185,19 @@ func ExtractFunctionStyleCommentTags(marker string, tagNames []string, lines []s
...
 
+// Tag represents a single comment tag.
+type Tag struct {
+	// Name is the name of the tag with no arguments.
+	Name string
+	// Args is a list of optional arguments to the tag.
+	Args []string
+	// Value is the value of the tag.
+	Value string
+}
```
jpbetz, line 196: This is the old tag representation.  Kept for backward compat

v2/comments.go
```diff
@@ -249,12 +246,17 @@ type TypedTag struct {
...
 type Arg struct {
-	Name  string
+	// Name is the name of a named argument. This is zero-valued for positional arguments.
+	Name string
+	// Value is the string value of an argument. It has been validated to match the Type.
 	Value string
-	Typ   ArgType
+	// Type is the type of the Value.
+	Type ArgType
 }
```
jpbetz, line 257: Is this string + type representation sufficient?  We could take this further and have types like:

```go
type String string

type (String) Type() ArgType {
  return TypeString
}
```

@aaron-prindle for thoughts on processing this downstream.

v2/argsparser.go
thockin: Do we have a use-case for it?

jpbetz: We have to somehow support `+k8s:<name>`. My test case for such tags was failing with the grammar.

v2/comments.go
```diff
@@ -213,6 +239,30 @@ type Tag struct {
...
+	Value string
+}
+
+type Arg struct {
+	Name  string
+	Value string
+	Typ   ArgType
+}
+
+type ArgType string
```
thockin: godoc

v2/argsparser.go
thockin: Do we have a use-case?  If we negative, do we also need '+' ?

Also I would write this as:

```
case r == '-' || r == '+' || unicode.IsDigit(r):
    buf.WriteRune(r)
    st = stNumber
```

...just to save on space

v2/argsparser.go
thockin: Alternate approach, accumulate all bytes, and when you hit the end of the loop `if st == stValue` then save that and return it.  Downside: any trailing comment is still there.  If someone calls this func again, it will "just work"

v2/comments.go
```diff
@@ -105,12 +105,11 @@ func ExtractSingleBoolCommentTag(marker string, key string, defaultVal bool, lin
...
 //   - 'marker' + "key(arg)=value"
 //   - 'marker' + "key(`raw string`)=value"
-//   - 'marker' + "key({"k1": "value1"})=value"
 //
-// The arg is optional.  It may be a Go identifier, a raw string literal
-// enclosed in back-ticks, or an object or array represented with a subset of
-// JSON syntax. If not specified (either as "key=value" or as
-// "key()=value"), the resulting Tag will have an empty Args list.
+// The arg is optional. It may be a Go identifier, a raw string literal enclosed
+// in back-ticks, an integer (with support for hex, octal or binary notation) or
```
thockin, line 111: back-ticks or doubel-quotes

v2/comments.go
```diff
@@ -251,32 +268,49 @@ type Arg struct {
...
+type Int int
+
+func (Int) Type() Type {
+	return TypeInt
 }
 
-// ArgType is the type of an arg.
-type ArgType string
+func (i Int) String() string {
+	return strconv.Itoa(int(i))
```
thockin, line 291: Why are you going thru the effort to parse string->int and then return it as a string anyway?

I'm not sure I see the value of the extra abstraction here

jpbetz: It made more sense when all of the types were backed by a string (`type Bool string`) :(

Removing.

v2/comments.go
```diff
@@ -178,26 +197,33 @@ func ExtractFunctionStyleCommentTags(marker string, tagNames []string, lines []s
...
 		line = strings.TrimSpace(line)
 		if !strings.HasPrefix(line, marker) {
 			continue
 		}
 		line = line[len(marker):]
-		s := initTagKeyScanner(line)
 
-		name, args, err := s.parseTagKey(tagNames)
+		line = stripTrailingComment(line)
+		kv := strings.SplitN(line, "=", 2)
```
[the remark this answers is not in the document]
jpbetz: Since a value can be another tag, or opaque text.  I think the only fix for trailing comments is to make the caller deal with it.. I'll drop the trailing comment removal.

I'll have the parser consume everything else and return it.

v2/argsparser.go
jpbetz: Agreed.  I'll leave the trailing comment in the value and parse it as opaque text.  See below comments for more detail.

v2/argsparser.go
jpbetz: Don't need negatives today, but since we are handling hex and octal, I threw it in for completeness.

Added + support.

jpbetz:
Thanks @thockin.  Excellent feedback.  I've applied it. I think this is getting close now.

I'm not happy with having to drop trailing comment support, but I think it's better to drop it for now and only add support if we can come up with a clean solution.

jpbetz:
@thockin All feedback applied.  This is now sufficient for extracting and parsing comments and also parsing nested tags that don't have a marker/group prefix ("+k8s:").

v2/comments.go
```diff
@@ -229,227 +225,185 @@ func (t Tag) String() string {
...
-	case '[':
-		return s.chompJSONFlavoredArray()
-	case '{':
-		return s.chompJSONFlavoredObject()
-	}
+		nameEnd := len(line)
+		if idx := strings.IndexAny(line, "(="); idx > 0 {
+			nameEnd = idx
+		}
+		ident := TagIdentifierFromString(line[:nameEnd])
```
thockin, line 260: If we keep "group", we could just move this to the state-machine -- would it be easier?

v2/comments.go
```diff
@@ -229,227 +225,185 @@ func (t Tag) String() string {
...
+}
 
-	switch s.PeekSkipSpaces() {
-	case ',':
-		s.Scan() // Consume ,
-		_, entriesEnd, err := s.chompJSONFlavoredObjectEntries()
-		if err != nil {
-			return 0, 0, err
+// ExtractAndParseTagWithArgs combines ExtractTags and ParseTagWithArgs.
+func ExtractAndParseTagWithArgs(marker string, group *string, lines []string) (map[TagIdentifier][]TypedTag, error) {
```
thockin, line 323: cute composition, but how much of it needs to be public API?

thockin (commented):
I wonder if we should make a sub-pkg for this now?

We tend to use "tags" which could be a good pkg name, or "directive" ?

Then you can pick a good name for types and functions, giving the old versions less nice names, and leaving a forwarding function from the root pkg.

gengo/v2/tags.ExtractCommentTags is `var ExtractCommentTags = tags.ExtractLegacySimpleTags` :)

v2/argsparser.go
```diff
@@ -0,0 +1,272 @@
...
+				break parseLoop
+			}
+		case stTag:
+			switch {
+			case isIdentInterior(r):
+				tag.WriteRune(r)
+			case r == '(':
+				incomplete = true
+				st = stArg
+			case r == '=':
```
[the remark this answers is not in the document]
jpbetz: I've added back trailing comment support. It's only allowed if there is no value part (for soundness of the parser).

v2/comments.go
jpbetz: I've dropped the group concept and allowed ":" in tag names.

v2/comments.go
jpbetz: Dropped it.  I kept a `ParseAll` utility. Let me know if even that is too much.

v2/codetags/types.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// Bool returns the boolean value of the argument.
+func (a BoolValue) Bool() bool {
+	return bool(a)
+}
+
+func (a BoolValue) Type() Type {
+	return TypeBool
+}
```
thockin, line 147: I'll be honest, I still don't see the point of the abstraction here, but I am not going to argue it further.

This just feels like a fancy way of saying:

```
type ArgType string

const (
    TypeInt ArgType = "int"
    TypeBool ArgType = "bool"
    TypeString ArgType = "string"
)

type Arg stuct {
    Name string
    Type ArgType
    Value string
}
```

...without offering a whole lot more safety.

v2/codetags/types.go
```diff
@@ -0,0 +1,147 @@
...
+	if len(a.Name) > 0 {
+		return fmt.Sprintf("%s: %v", a.Name, a.Value)
+	}
+	return fmt.Sprintf("%v", a.Value)
+}
+
+// TypedValue represents a parsed value.
+// It may be a string, int, or bool. TypedValue retains the original string representation
+// of the value that was parsed from the tag.
+type TypedValue interface {
```
thockin, line 75: TypedValue makes it sound like the value of the tag, not the value of the arg.  See my other comment about abstraction.  Naming things is hard and you have made two things to name instead of just one.

v2/codetags/parser.go
```diff
@@ -0,0 +1,372 @@
...
+				incomplete = false
+				st = stTrailingComment
+			default:
+				break parseLoop
+			}
+		case stTrailingComment:
+			i = len(runes) - 1
+			break parseLoop
+		default:
+			panic("unknown state")
```
thockin, line 350: I know it's kind an `ASSERT()` but it still feels dirty to panic.

And anyway, you could add the string of the unhandled state :)

v2/comments.go
```diff
@@ -74,6 +73,10 @@ func ExtractCommentTags(marker string, lines []string) map[string][]string {
 // If the tag is not found, the default value is returned.  Values are asserted
 // to be boolean ("true" or "false"), and any other value will cause an error
 // to be returned.  If the key has multiple values, the first one will be used.
+//
+// This function is a wrapper around codetags.Extract and codetags.Parse, but only supports tags with
+// a single position arg of type string, and a value of type bool.
+// This function uses the full group:name identifier as the returned Tag.name.
```
thockin, line 79: references "group"

v2/comments.go
```diff
@@ -92,117 +95,57 @@ func ExtractSingleBoolCommentTag(marker string, key string, defaultVal bool, lin
...
-//			}, {
-//				Name: "buzz",
-//				Args: []string{"{\"a\": 1, \"b\": \"x\"}"},
-//				Value: ""
-//		}}
-//
-// This function should be preferred instead of ExtractCommentTags.
+// This function is a wrapper around codetags.Extract and codetags.Parse, but only supports tags with
+// a single position arg of type string.
+// This function uses the full group:name identifier as the returned Tag.name.
```
thockin, line 102: references "group"

v2/codetags/extractor.go
```diff
@@ -0,0 +1,60 @@
...
+//	+k8s:format=k8s-long-name
+//
+// This function treats all characters before the first '=', '(' or end-of-line as the name
+// of the tag.
+//
+// Then this function will return:
+//
+//	map[string][]string{
+//		"required": {"required"},
+//		"required": {"format=k8s-long-name"},
```
thockin, line 40: why is the key here "required" ?

v2/codetags/extractor.go
```diff
@@ -0,0 +1,60 @@
...
+	"strings"
+)
+
+// Extract extracts comments for special metadata tags. Lines are filtered by a
+// string prefix that matching lines must start with.
+//
+// Example: When called with prefix "+k8s:", lines:
+//
+//	Comment line without marker
+//	+k8s:required
```
thockin, line 29: Can you include a (fake) example which explains why it returns `[]string` ?

thockin: Also mayber an example with args to show why the map-key is included in the value?  E.g. I think it would be somewthing like:

```
"withArgs": {"withArgs(abc123)=value"},
```

right?

v2/comments.go
```diff
@@ -46,7 +45,7 @@ import (
 //
 //	map[string][]string{"foo":{"value1, "value2"}, "bar": {""}, "baz": {`"qux"`}}
 //
-// Deprecated: Use ExtractFunctionStyleCommentTags.
+// Deprecated: Prefer codetags.Extract, codetags.Parse and codetags.ExtractAndParse.
```
thockin, line 48: ExtractAndParse no longer exists?

thockin (commented):
mostly small things

v2/codetags/extractor.go
jpbetz: SG.  I found a bug in Extract when expanding the test coverage. Fixing that too.  The example now covers this functions capabilities more extensively.

v2/codetags/extractor.go
jpbetz: copy-pasta.  Fixing.

v2/codetags/types.go
jpbetz: Apologies.  I'be been searching for ways to minimize the amount of parsing callers need to do.  This feedback suggests I've over-fit for that. I'll back this out.

jpbetz:
@thockin Feedback applied.  I've reverted back to the type field approach.  I found a bug in extract so I've cleaned that code up. Most of the rest of the work is test and godoc cleanup.

v2/codetags/types.go
thockin: I feel like I have come full circle in my Go journey - I initially looked for places to build abstract and I have now reached a point where I am trying to take it out. :)

v2/codetags/extractor.go
```diff
@@ -49,12 +60,29 @@ func Extract(prefix string, lines []string) map[string][]string {
 		line = line[len(prefix):]
 
 		// Find the end of the presumed tag name.
-		nameEnd := len(line)
-		if idx := strings.IndexAny(line, "(="); idx > 0 {
-			nameEnd = idx
+		nameEnd := findNameEnd(line)
+		if nameEnd == 0 {
+			continue
```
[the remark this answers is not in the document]
jpbetz: Keeping it SGTM. Updated.  I've also refined the whitespace handling to preserve the value as much as possible. This feels inline with our goal of simplifying these functions down to a single purpose.

jpbetz:
@thockin feedback applied

v2/codetags/parser_test.go
```diff
@@ -17,22 +17,32 @@ limitations under the License.
...
 
 import (
-	"fmt"
 	"testing"
 )
 
-func TestParseTagKeyWithTagNames(t *testing.T) {
-	mkss := func(s ...string) []string { return s }
+func TestParse(t *testing.T) {
+	mkss := func(s ...string) []Arg {
```
yongruilin, line 24: nit: mkss -> mksa ?

v2/codetags/parser_test.go
```diff
@@ -45,31 +55,72 @@ func TestParseTagKeyWithTagNames(t *testing.T) {
 		{input: "name(has-dash)", expectKey: "name", err: true},
 		{input: "name(multiple, args)", expectKey: "name", err: true},
 		{input: "name(noClosingParen", expectKey: "name", err: true},
+
+		// invalid args
+		{input: "name(arg1, arg2)", err: true},
+		{input: "argWithDash(arg-name) ", err: true},
+		{input: "withRaw(`a = b`)", expectKey: "withRaw", expectArgs: mkss("a = b")},
```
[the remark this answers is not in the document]
jpbetz: Updated

jpbetz:
/hold
For @thockin LGTM for merge.

thockin:
FWIW, we should probably fix any places that accept payloads and clarify that the args are strings.  No more naked JSON.

jpbetz:
> FWIW, we should probably fix any places that accept payloads and clarify that the args are strings.  No more naked JSON.

We haven't accumulated any uses in k/k yet (I just double checked).  The only use is in the validation-gen dev branch and we can easily fix.  We'll follow up in the clarification in godoc too.

v2/codetags/parser.go
```diff
@@ -47,56 +47,78 @@ type tagKey struct {
...
+//	"name" // no value
+//	"name=value"
+//	"name=values include all the content after the = sign including any special characters (@*#^&...)"
+//
+// Comments are treated as part of the value, if a value is present.
+//
+// For example,
+//
+//	"key // This tag has no value, so this comment is ignored"
+//	"key=value // Comments after values are treated as part of the value"
```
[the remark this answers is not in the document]
jpbetz: I didn't think about this long enough before responding.  Opening a convo on chat to hammer it down.

thockin:
I see a bunch of tests that use naked JSON with `validateFalse` ?

PR #303, opened 2025-05-28
Treat embedded fields as inline

### LINKED ISSUE #300: JSON tag "inline" is not a real thing
`k8s.io/gengo/v2/parser/tags.LookupJSON()` asserts that "inline" is a field in the `json` tag, but Go defines no such thing.

thockin:
Fixes #300

### DIFF
--- a/v2/parser/tags/json.go
+++ b/v2/parser/tags/json.go
@@ -41,6 +41,8 @@ func (t JSON) String() string {
 	if t.Omitempty {
 		tag += ",omitempty"
 	}
+	// "inline" isn't (yet) a standard json tag, but it is used by
+	// gengo to indicate that the field should be inlined.
 	if t.Inline {
 		tag += ",inline"
 	}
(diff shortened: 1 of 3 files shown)


thockin:
Ahh, good point.

Interestingly, I ran this against k/k and nothing changed wrt ObjectMeta.

> If I'm reading this right, we'd interpret this differently than the rest of Kubernetes if we make this change?

I think the fact that we treat embedded fields as their type-name is wrong.  We got this far because of `inline` but it's not strictly correct. 

Changed to consider it inline if embedded AND JSON name is "".

PTAL

jpbetz:
> Changed to consider it inline if embedded AND JSON name is "".

SGTM. This is unambigious.

thockin:
I can try to look at this later if you don't get it first. 

I think it's reasonable and consistent to put subfield on an embedded struct, but I could be argued. It's certainly a weird case.

PR #304, opened 2025-05-29
Extend named parser to also parse chained tags

jpbetz:
Builds on https://github.com/kubernetes/gengo/pull/299

Extends the grammar to parse values, support chained tags and ignore all trailing comments. This is opt-in for existing functions (for backward compat) and opt-out for the direct calls to the new parser API (to encourage new use cases to parse values).

For example,

```
// Chain tags with a dedicated operator:

name(limit: 10, path: "/xyz")=+anotherTag(size: 100)

// Values must be string, bool, int or identifier:

name=identifier
name="double-quoted value"
name=`backtick-quoted value`
"name(100)
name(true)
```

Change to grammar:

```
<tag>             ::= <tagName> [ "(" [ <args> ] ")" ] [ "=" <value> | "=+" <tag> ]
<value>           ::= <identifier> | <string> | <int> | <bool>
```

### DIFF
--- a/v2/codetags/parser.go
+++ b/v2/codetags/parser.go
@@ -142,17 +142,16 @@ func RawValues(enabled bool) ParseOption {
 	}
 }
 
-const (
-	stTag           = "stTag"
-	stMaybeArgs     = "stMaybeArgs"
-	stArg           = "stArg"
-	stArgEndOfToken = "stArgEndOfToken"
-	stMaybeValue    = "stMaybeValue"
-	stValue         = "stValue"
-	stMaybeComment  = "stMaybeComment"
-)
-
 func parseTag(input string, opts parseOpts) (Tag, error) {
+	const (
+		stTag           = "stTag"
+		stMaybeArgs     = "stMaybeArgs"
+		stArg           = "stArg"
+		stArgEndOfToken = "stArgEndOfToken"
+		stMaybeValue    = "stMaybeValue"
+		stValue         = "stValue"
+		stMaybeComment  = "stMaybeComment"
+	)
 	var startTag, endTag *Tag // both ends of the chain when parsing chained tags
 
 	// accumulators
@@ -216,7 +215,7 @@ func parseTag(input string, opts parseOpts) (Tag, error) {
 	var err error
 	st := stTag
 parseLoop:
-	for r := s.skipWhitespacePeek(); r != EOF; r = s.skipWhitespacePeek() {
+	for r := s.peek(); r != EOF; r = s.peek() {
 		switch st {
 		case stTag:
 			switch {
@@ -276,10 +275,14 @@ parseLoop:
 
 				switch {
 				case r == ',' || r == ')': // positional arg
+					if r == ',' {
+						r = s.skipWhitespace() // allow whitespace after ,
+					}
 					saveBoolOrString(identifier)
 					st = stArgEndOfToken
 				case r == ':': // named arg
-					s.next() // consume :
+					s.next()               // consume :
+					r = s.skipWhitespace() // allow whitespace after :
 					saveName(identifier)
 					st = stArg
 				default:
@@ -291,7 +294,8 @@ parseLoop:
 		case stArgEndOfToken:
 			switch {
 			case r == ',':
-				s.next() // consume ,
+				s.next()               // consume ,
+				r = s.skipWhitespace() // allow whitespace after ,
 				st = stArg
 			case r == ')':
 				s.next() // consume )
@@ -363,7 +367,7 @@ parseLoop:
 			}
 		case stMaybeComment:
 			switch {
-			case r == '/' && s.peekN(1) == '/':
+			case s.nextIsTrailingComment():
 				s.remainder()
 			default:
 				break parseLoop
--- a/v2/codetags/parser.go
+++ b/v2/codetags/parser.go
@@ -24,15 +24,20 @@ import (
 	"unicode"
 )
 
-// Parse parses a comment tag into a TypedTag, or returns an error if the tag
+// Parse parses a tag string into a TypedTag, or returns an error if the tag
 // string fails to parse.
 //
-// A tag consists of a name, optional arguments, and an optional value. For example,
+// Any enabled ParseOptions modify the behavior of the parser. The below
+// describes only the default behavior.
+//
+// A tag consists of a name, optional arguments, and an optional scalar value or
+// tag value. For example,
 //
 //	"name"
 //	"name=50"
 //	"name("featureX")=50"
 //	"name(limit: 10, path: "/xyz")=text value"
+//	"name(limit: 10, path: "/xyz")=+anotherTag(size: 100)"
 //
 // Arguments are optional and may be either:
 //   - A single positional argument.
@@ -63,30 +68,34 @@ import (
 // typically used first to find and isolate tag strings matching a specific
 // prefix. Those extracted strings can then be parsed using this function.
 //
-// The value part of the tag is optional and follows an equals sign "=".
-// If no equals sign and value are present, the `Value` field of the
-// resulting TypedTag will be an empty string.
+// The value part of the tag is optional and follows an equals sign "=". If a
+// value is present, it must be a string, int, boolean, identifier, or tag.
 //
 // For example,
 //
 //	"name" // no value
-//	"name=value"
-//	"name=values include all the content after the = sign including any special characters (@*#^&...)"
+//	"name=identifier"
+//	"name="double-quoted value""
+//	"name=`backtick-quoted value`"
+//	"name(100)"
+//	"name(true)"
+//	"name=+anotherTag"
+//	"name=+anotherTag(size: 100)"
 //
-// Comments are treated as part of the value, if a value is present.
+// Trailing comments are ignored unless opts.ParseValues=false, in which case they
+// are treated as part of the value. See ParseValues.RawValues for details.
 //
 // For example,
 //
-//	"key // This tag has no value, so this comment is ignored"
-//	"key=value // Comments after values are treated as part of the value"
+//	"key=value // This comment is ignored"
 //
 // Formal Grammar:
 //
-// <tag>             ::= <tagName> [ "(" [ <args> ] ")" ] [ "=" <tagValue> ]
-// <args>            ::= <argValue> | <namedArgs>
+// <tag>             ::= <tagName> [ "(" [ <args> ] ")" ] [ ( "=" <value> | "=+" <tag> ) ]
+// <args>            ::= <value> | <namedArgs>
 // <namedArgs>       ::= <argNameAndValue> [ "," <namedArgs> ]*
-// <argNameAndValue> ::= <identifier> ":" <argValue>
-// <argValue>        ::= <identifier> | <string> | <int> | <bool>
+// <argNameAndValue> ::= <identifier> ":" <value>
+// <value>           ::= <identifier> | <string> | <int> | <bool>
 //
 // <tagName>       ::= [a-zA-Z_][a-zA-Z0-9_-.:]*
 // <identifier>    ::= [a-zA-Z_][a-zA-Z0-9_-.]*
@@ -95,17 +104,24 @@ import (
 // <int>           ::= /* Standard Go integer literals (decimal, 0x hex, 0o octal, 0b binary),
 // ...                    with an optional +/- prefix. */
 // <bool>          ::= "true" | "false"
-// <tagValue>      ::= /* All text following the "=" sign to the end of the string. */
-func Parse(tagText string) (TypedTag, error) {
-	tagText = strings.TrimSpace(tagText)
-	return parseTagKey(tagText)
+func Parse(tag string, opts ParseOptions) (TypedTag, error) {
+	tag = strings.TrimSpace(tag)
+	return parseTag(tag, opts)
+}
+
+// ParseOptions controls the behavior of Parse.
+type ParseOptions struct {
+	// RawValues disables parsing of the value part of the tag. If true, the Value
+	// field will contain all text following the "=" sign, up to the last
+	// non-whitespace character, and ValueType will be set to ValueTypeRaw.
+	RawValues bool
 }
 
 // ParseAll calls Parse on each tag in the input slice.
-func ParseAll(tags []string) ([]TypedTag, error) {
+func ParseAll(tags []string, opts ParseOptions) ([]TypedTag, error) {
 	var out []TypedTag
 	for _, tag := range tags {
-		parsed, err := Parse(tag)
+		parsed, err := Parse(tag, opts)
 		if err != nil {
 			return nil, err
 		}
@@ -115,26 +131,41 @@ func ParseAll(tags []string) ([]TypedTag, error) {
 }
 
 const (
-	stBegin           = "stBegin"
-	stTag             = "stTag"
-	stArg             = "stArg"
-	stNumber          = "stNumber"
-	stPrefixedNumber  = "stPrefixedNumber"
-	stQuotedString    = "stQuotedString"
-	stNakedString     = "stNakedString"
-	stEscape          = "stEscape"
-	stEndOfToken      = "stEndOfToken"
-	stMaybeValue      = "stMaybeValue"
-	stValue           = "stValue"
+	stBegin = "stBegin"
+	stTag   = "stTag"
+	stArg   = "stArg"
+
+	// arg value parsing states
+	stArgNumber         = "stArgNumber"
+	stArgPrefixedNumber = "stArgPrefixedNumber"
+	stArgQuotedString   = "stArgQuotedString"
+	stArgNakedString    = "stArgNakedString"
+	stArgEscape         = "stArgEscape"
+	stArgEndOfToken     = "stArgEndOfToken"
+
+	// tag value parsing states
+	stMaybeValue          = "stMaybeValue"
+	stValue               = "stValue"
+	stValueTagOrNumber    = "stValueTagOrNumber"
+	stValueNumber         = "stValueNumber"
+	stValuePrefixedNumber = "stValuePrefixedNumber"
+	stValueQuotedString   = "stValueQuotedString"
+	stValueNakedString    = "stValueNakedString"
+	stValueEscape         = "stValueEscape"
+
 	stMaybeComment    = "stMaybeComment"
 	stTrailingSlash   = "stTrailingSlash"
 	stTrailingComment = "stTrailingComment"
 )
 
-func parseTagKey(input string) (TypedTag, error) {
+func parseTag(input string, opts ParseOptions) (TypedTag, error) {
+	var startTag, endTag *TypedTag // both ends of the chain when parsing chained tags
+
 	tag := bytes.Buffer{}   // current tag name
 	args := []Arg{}         // all tag arguments
 	value := bytes.Buffer{} // current tag value
+	var valueType ValueType // current value type
+	var hasValue bool       // true if the tag has a value
 
 	cur := Arg{}          // current argument accumulator
 	buf := bytes.Buffer{} // string accumulator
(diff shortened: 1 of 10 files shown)


v2/codetags/types.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// ValueTag is value of the tag when the tag's value is another tag.
+	// Only valid when ValueType is ValueTypeTag.
+	ValueTag *TypedTag
+
+	// ValueType is the type of the value.
+	// If ValueType is ValueTypeTag, ValueTag is set.
+	// If ValueType is ValueTypeString, ValueTypeBool or ValueTypeInt, Value is the string representation of the value.
+	// If ValueType is empty, the tag has no value.
+	ValueType ValueType
```
jpbetz, line 42: Alternative's I've considered:

- Leave chained tags as string values in `Value` instead of providing a field to access the parsed tag.
- Change `Parse()` to return a slice of `TypedTag` where each item in the slice represents a chained tag.
- Change `Value` to be either `any` or an interface that can either be a `string` or a `TypedTag`.

jpbetz:
@thockin This adds typed value parsing.  I've opened it as a separate PR for now, since it's not clear if we want to merge https://github.com/kubernetes/gengo/pull/299 or combine this with it yet.  Happy to consolidate once this has gotten eyes.

Only the last commit needs review.

thockin (commented):
I have to attention shift but wanted to send commentsbefore I do.

Given how duplicative this is, maybe we should look at a slightly different form of managing states?  Something like:

```
type runeQueue struct {
    buf []rune
    head int // index into buf of the next rune
}

func (rq *runeQueue) Pop() rune {
    return rq.buf[rq.cur++]
}
func (rq *runeQueue) Peek() rune {
    return rq.PeekN(0)
}
func (rq *runeQueue) PeekN(n int) rune {
    return rq.buf[rq.cur+n]
}
```

Then always operate on `Peek()` until you know you want to consume the rune.

Then you can turn sub-machines into functions, like:

```
// Precondition: rq is assumed to be at the beginning of a number token
// Postcondition: rq is at the next rune after the number
func parseNumber(rq *runeQueue) (string, error) {
    buf bytes.Buffer{}
    switch {
        // handle all number states
    }
    return buf.String()
}
````

And the "outer" FSM recognizes a digit, '+' + digit, or '-' as a number and calls `parseNumber()`. When that returns you can look at the next character and decide what the next state is, or if it is an error.  You still need the outer-FSM's context to know if the next thing is allowed, but you can encapsulate the logic of parsing a number.

I'll see if I can clarify later.

File: v2/codetags/parser.go [hunk 2]
```diff
@@ -324,14 +393,102 @@ parseLoop:
...
+				value.WriteRune(r)
+				st = stValueNumber
+			case isIdentBegin(r):
+				if err := saveTag(); err != nil {
+					return TypedTag{}, err
+				}
+				incomplete = false
+				st = stMaybeValue
+				tag.WriteRune(r)
+				st = stTag
```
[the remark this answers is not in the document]
jpbetz: `saveTag()` does the dirty work. It tracks the previously parsed tag key. And then `st = stTag`  initiates parsing of the next tag.  At the very end of the FSM there is a final call to `saveTag()` to track the last tag, followed by a call to `saveValue()` to set a value on the last tag, if there is one.

This MIGHT be easier to follow in the new code that uses the next() and peek()  operations you suggested.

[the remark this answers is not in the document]
jpbetz: (`parser_test.go` tests to a depth of 4 nested tags.)

jpbetz:
@thockin micro-FSMs split out for the string, number and identity parsing.  I tried to keep it simple and forklift the states we already had into the micro-FSMs.  It seems to be a win.  We're close to a slippery slope... next logical step would be to write a lexer and then parse based on lexer tokens.  If we go much farther, we're going to reinvent text/scanner :rofl:

v2/codetags/parser.go
```diff
@@ -183,139 +206,142 @@ func parseTagKey(input string) (TypedTag, error) {
...
+			}
+		case stMaybeArgs:
+			switch {
 			case r == '(':
+				s.next()
 				incomplete = true
 				st = stArg
 			case r == '=':
+				s.next()
+				hasValue = true
```
thockin, line 281: should this set incomplete?  If you hit EOF right after `=` it should be an error, right?

v2/codetags/parser.go
```diff
@@ -183,139 +206,142 @@ func parseTagKey(input string) (TypedTag, error) {
...
+				s.next()
 				st = stMaybeComment
 			default:
 				break parseLoop
 			}
 		case stArg:
 			switch {
 			case unicode.IsSpace(r):
+				s.next()
 				continue
```
thockin, line 293: We should probably get rid of explicit `continue` unless there are no other statements.  Or even then, just leave a comment?

v2/codetags/parser.go
```diff
@@ -183,139 +206,142 @@ func parseTagKey(input string) (TypedTag, error) {
...
+	for r := s.peek(); r != EOF; r = s.peek() {
 		switch st {
 		case stBegin:
 			switch {
 			case unicode.IsSpace(r):
+				s.next()
 				continue
 			case isIdentBegin(r):
-				tag.WriteRune(r)
 				st = stTag
```
thockin, line 257: Should we add comments like `// do not consume` ?

v2/codetags/scanner.go
```diff
@@ -0,0 +1,205 @@
...
+	for r := s.peek(); r != EOF; r = s.peek() {
+		switch st {
+		case stBegin:
+			switch {
+			case r == '+':
+				s.next() // consume +
+			case r == '0':
+				buf.WriteRune(s.next())
+				st = stPrefixNumber
+			case r == '-' || unicode.IsDigit(r):
```
thockin, line 79: as mentioned in chat, we need to allow `+0x1234` and `-0b1111011` sorts of things, so we need an intermediate state like:

```
        switch st {    
        case sBegin:    
            switch {    
            case r == '0':    
                buf.WriteRune(s.next())    
                st = stPrefix    
            case r == '-' || r == '+':    
                buf.WriteRune(s.next())    
                st = stPosNeg    
            case unicode.IsDigit(r):    
                buf.WriteRune(s.next())    
                st = stNumber    
            default:    
                break parseLoop    
            }    
        case stPosNeg:    
            switch {    
            case r == '0':    
                buf.WriteRune(s.next())    
                st = stPrefix    
            case unicode.IsDigit(r):    
                buf.WriteRune(s.next())    
                st = stNumber   
            default:    
                break parseLoop    
            }    
        case stPrefix:    
            switch {    
            case unicode.IsDigit(r):    
                buf.WriteRune(s.next())    
                st = stNumber    
            case r == 'x' || r == 'o' || r == 'b':    
                buf.WriteRune(s.next())    
                st = stNumber    
            default:    
                break parseLoop    
            }    
        case stNumber:                                                                                                                  
            const hexits := "abcdefABCDEF"    
            switch {    
            case unicode.IsDigit(r) || strings.Contains(hexits, string(r)):    
                buf.WriteRune(s.next())    
            default:    
                break parseLoop    
            }    
        }    
```

v2/codetags/scanner.go
```diff
@@ -0,0 +1,205 @@
...
+				st = stNumber
+			case r == 'x' || r == 'o' || r == 'b':
+				incomplete = true
+				buf.WriteRune(s.next())
+				st = stNumber
+			default:
+				break parseLoop
+			}
+		case stNumber:
+			hexits := "abcdefABCDEF"
```
thockin, line 98: `const hexits`

v2/codetags/parser.go
```diff
@@ -183,139 +206,142 @@ func parseTagKey(input string) (TypedTag, error) {
...
-				saveName()
-				st = stArg
+				st = stArgEndOfToken
+			case isIdentBegin(r):
+				str, err := s.nextIdent(isIdentInterior)
+				if err != nil {
+					return TypedTag{}, err
+				}
+				buf.WriteString(str)
+				r = s.peek()
```
thockin, line 322: this could do with a small comment

v2/codetags/parser.go
```diff
@@ -324,19 +350,59 @@ parseLoop:
 		case stMaybeValue:
 			switch {
 			case r == '=':
+				s.next()
+				hasValue = true
```
thockin, line 354: Same as elsewhere: should this set incomplete? If you hit EOF right after = it should be an error, right?

v2/codetags/parser.go
```diff
@@ -324,19 +350,59 @@ parseLoop:
...
 			case unicode.IsSpace(r):
+				s.next()
+				st = stMaybeComment
+			default:
+				break parseLoop
+			}
+		case stValue:
+			switch {
+			case opts.rawValues: // When enabled, consume all remaining chars
+				value.WriteRune(s.next())
```
thockin, line 365: `s.remainder()` ?

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			input:  "name(lower)",
+			expect: mkta("name", mksa("lower")),
+		},
+		{
+			name:   "uppercase argument",
+			input:  "name(CAPITAL)",
+			expect: mkta("name", mksa("CAPITAL")),
+		},
+		{
+			name:   "mixed case argument 1",
```
thockin, line 130: You also have "mixed case argument" above

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			input:  "name(MiXeD)",
+			expect: mkta("name", mksa("MiXeD")),
+		},
+		{
+			name:   "mixed case argument 2",
+			input:  "name(mIxEd)",
+			expect: mkta("name", mksa("mIxEd")),
+		},
+		{
+			name:   "underscore in argument",
```
thockin, line 140: how about just "_" as the arg?

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			input:  `name("hasQuotes")`,
+			expect: mkta("name", mksa("hasQuotes")),
+		},
+		{
+			name:   "raw quoted argument",
+			input:  "name(`hasRawQuotes`)",
+			expect: mkta("name", mksa("hasRawQuotes")),
+		},
+		{
+			name:   "raw string with spaces",
```
thockin, line 155: how about a test for `"true"` and `"false"` (maybe do one with backticks) to distinguish `string("true")` from `bool(true)` ?

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+		},
+
+		// Error cases for arguments
+		{
+			name:  "space in argument",
+			input: "name(has space)",
+			err:   true,
+		},
+		{
+			name:  "multiple comma-separated args",
```
thockin, line 167: this is because they are positional?  Maybe say that

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			input: "name(multiple, args)",
+			err:   true,
+		},
+		{
+			name:  "unclosed parenthesis",
+			input: "name(noClosingParen",
+			err:   true,
+		},
+		{
+			name:  "comma-separated args",
```
thockin, line 177: dup with "multiple comma-separated args" case above?

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			input: "name(arg1, arg2)",
+			err:   true,
+		},
+		{
+			name:  "unclosed raw string",
+			input: "badRaw(missing`)",
+			err:   true,
+		},
+		{
+			name:  "mixed arg formats",
```
thockin, line 187: not sure what this is proving?

Just making it "err: true" means you can't see WHY it failed.  Maybe substring match?

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			}),
+		},
+
+		// Numeric argument tests
+		{
+			name:  "numeric arguments",
+			input: "numbers(n1: 2, n2: -5, n3: 0xFF00B3, n4: 0o04167, n5: 0b10101)",
+			expect: mkta("numbers", []Arg{
+				{Name: "n1", Value: "2", Type: ArgTypeInt},
+				{Name: "n2", Value: "-5", Type: ArgTypeInt},
```
thockin, line 236: +5

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+				{Name: "n3", Value: "0xFF00B3", Type: ArgTypeInt},
```
thockin, line 237: -0x1234
+0x...
-0b...
+0b...
0...
-0...
+0...

v2/codetags/parser_test.go
```diff
@@ -29,100 +30,462 @@ func TestParse(t *testing.T) {
...
+			name:  "3 level nested tag values with comment",
+			input: `key=+key2=+key3 // comment`,
+			expect: mktt("key", &TypedTag{
+				Name:      "key2",
+				ValueType: ValueTypeTag,
+				ValueTag:  &TypedTag{Name: "key3"},
+			}),
+		},
+		{
+			name:  "4 level nested tag values with comment",
```
thockin, line 418: ...and args?

v2/codetags/scanner_test.go
```diff
@@ -0,0 +1,324 @@
...
+			name:  "positive integer with plus sign",
+			input: "+123",
+			want:  "123",
+		},
+		{
+			name:  "zero",
+			input: "0",
+		},
+		{
+			name:  "hex number",
```
thockin, line 111: pos/neg hex, octal, bin

v2/codetags/scanner_test.go
```diff
@@ -0,0 +1,324 @@
...
+			input:   "`hello",
+			wantErr: true,
+		},
+		{
+			name:    "invalid escape sequence",
+			input:   `"hello \n world"`,
+			wantErr: true,
+		},
+		{
+			name:  "string followed by other content",
```
thockin, line 217: cases for `"true"` and `"false"`

v2/codetags/types.go
```diff
@@ -17,20 +17,32 @@ limitations under the License.
 package codetags
 
 import (
-	"fmt"
+	"strconv"
 	"strings"
 )
 
 // TypedTag represents a single comment tag with typed args.
 type TypedTag struct {
```
thockin, line 25: For consideration: does "Typed" add much here?  Should this be Tag or should Arg be TypedArg ?

v2/codetags/parser.go
```diff
@@ -114,45 +126,56 @@ func ParseAll(tags []string) ([]TypedTag, error) {
...
 	stMaybeComment    = "stMaybeComment"
 	stTrailingSlash   = "stTrailingSlash"
 	stTrailingComment = "stTrailingComment"
 )
 
-func parseTagKey(input string) (TypedTag, error) {
+func parseTag(input string, opts parseOpts) (TypedTag, error) {
+	var startTag, endTag *TypedTag // both ends of the chain when parsing chained tags
+
 	tag := bytes.Buffer{}   // current tag name
```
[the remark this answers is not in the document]
jpbetz: +1.  I've organized, renamed and simplified the vars.

v2/codetags/parser.go
```diff
@@ -183,139 +206,142 @@ func parseTagKey(input string) (TypedTag, error) {
...
+		return nil
+	}
+	saveValue := func() {
+		endTag.Value = value.String()
+		if opts.rawValues {
+			endTag.ValueType = ValueTypeRaw
+			return
+		}
+		endTag.ValueType = valueType
+		if valueType == ValueTypeString && (endTag.Value == "true" || endTag.Value == "false") {
```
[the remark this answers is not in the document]
jpbetz: Oops, fixed.

v2/codetags/scanner.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (s *scanner) peek() rune {
+	if s.pos >= len(s.buf) {
+		return EOF
+	}
+	return s.buf[s.pos]
+}
+
+func (s *scanner) peekN(n int) rune {
+	if s.pos >= len(s.buf) {
```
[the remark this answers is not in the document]
jpbetz: Turns out I never needed peekN.  Removing.

v2/codetags/scanner.go
```diff
@@ -0,0 +1,205 @@
...
+		return EOF
+	}
+	r := s.buf[s.pos]
+	s.pos++
+	return r
+}
+
+func (s *scanner) peek() rune {
+	if s.pos >= len(s.buf) {
```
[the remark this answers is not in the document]
jpbetz: I removed peekN.

v2/codetags/parser.go
jpbetz: I'll add the check for typed values.  For ParseRaw, I'll allow an empty string value.

v2/codetags/parser.go
```diff
@@ -183,139 +206,142 @@ func parseTagKey(input string) (TypedTag, error) {
...
+		}
+	}
 	st := stBegin
 parseLoop:
-	for i, r = range runes {
+	for r := s.peek(); r != EOF; r = s.peek() {
 		switch st {
 		case stBegin:
 			switch {
 			case unicode.IsSpace(r):
```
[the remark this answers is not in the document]
jpbetz: Added.

v2/codetags/parser.go
jpbetz: Fixed and added test.  For ParseRaw values it remains allowed.

v2/codetags/parser_test.go
jpbetz: Dropped this test.  Added substring error matches.

v2/codetags/types.go
jpbetz: After moving to a dedicated package, `Tag` is sufficient.  There was another Tag type in root package that this started in.

Fixed.

jpbetz:
Feedback applied

v2/codetags/parser.go
```diff
@@ -288,26 +273,28 @@ parseLoop:
 					return Tag{}, err
 				}
 				r = s.peek() // reset r after nextIdent
+
 				switch {
-				case r == ',' || r == ')' || s.skipWhitespace():
+				case r == ',' || r == ')': // positional arg
```
thockin, line 278: this gets complicated with whitespace:

```
// +k8s:tag(foo    ) // should be OK?
// +k8s:tag(  n1:  foo  ,  n2:  bar  ) // should be OK?
// +k8s:tag(foo : bar) // should be an error?
```

v2/codetags/parser.go
```diff
@@ -371,27 +363,11 @@ parseLoop:
...
-			default:
-				break parseLoop
-			}
-		case stTrailingSlash:
-			switch {
-			case r == '/':
-				s.next()
-				incomplete = false
-				st = stTrailingComment
+			case r == '/' && s.peekN(1) == '/':
```
thockin, line 366: We could switch tag line-comments to use `#` instead - better or worse?

v2/codetags/parser.go
```diff
@@ -143,16 +143,13 @@ func RawValues(enabled bool) ParseOption {
 }
 
 const (
```
thockin, line 145: could be inside parseTag() ?

v2/codetags/parser.go
```diff
@@ -216,56 +213,44 @@ func parseTag(input string, opts parseOpts) (Tag, error) {
 		endTag.Value = value
 		endTag.ValueType = valueType
 	}
-	st := stBegin
+	var err error
+	st := stTag
 parseLoop:
-	for r := s.peek(); r != EOF; r = s.peek() {
+	for r := s.skipWhitespacePeek(); r != EOF; r = s.skipWhitespacePeek() {
```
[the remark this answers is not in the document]
thockin: Also - test cases

v2/codetags/parser.go
jpbetz: I prefer it.  I find the use of `//` as a comment tag inside other `//` comments :exploding_head: 

We did support `//` before this change, but only for a few months.  Are we okay to break?  If so I'm happy to make the change.

thockin: I think a break is OK

jpbetz:
Switch to # for comments pushed.

jpbetz:
/hold
 Something about this most recent iteration does not work with downstream tags

v2/codetags/parser.go
```diff
@@ -63,30 +66,34 @@ import (
...
 //
-// Comments are treated as part of the value, if a value is present.
+// Trailing comments are ignored unless the RawValues option is enabled, in which
+// case they are treated as part of the value.
 //
 // For example,
 //
-//	"key // This tag has no value, so this comment is ignored"
-//	"key=value // Comments after values are treated as part of the value"
+//	"key=value // This comment is ignored"
```
thockin, line 88: Update comments about comments to '#'

thockin (commented):
LGTM, just minor things.

v2/codetags/parser.go
jpbetz: Thanks. I swept all godoc.

jpbetz:
Added two small accessor utilities (`Tag.NamedArg(string) (Arg, bool)` and `Tag.PositionalArg() (Arg, bool)`) based on trying to use this code downstream.

thockin:
I can't find anything else to complain about.  I am sure we missed something but this looks pretty clean to me.

/lgtm
/approve

Issue #298, opened 2025-05-23
`PointerTo()` produces different results for repeat invocations


thockin:
For every type in a Universe, there is a single `Type` object instance.  This allows comparisons, e.g. to `types.String`.  The way `PointerTo()` is implement, `PointerTo(String) != PointerTo(String)`.

We should probably change it to do be safer.

thockin:
Also: if we want other common types like `error` or utils like `SliceOf()` or `ArrayOf()`...

thockin:
You are right, for the few places that call `PointerTo()`, we don't have a reference to the `Universe`, so we can't "just" change it to `u.PointerTo(t)`.

We could maybe do something like:
* store `*Universe` in each `Type`
* call `t.Pointer()` which would:
  - lookup the pointer type in the Universe
  - add it if needed
  - return it

This is super racy, since all the universe setup is assumed to be single-threaded (no mutexes or anything, just a map).  So I think we can't do that.  I hate the Unverse API for this reason but changing it is a lot of work and risk.

We could instead keep a cache of synthetic types (struct with a mutex and a Package):
* store `*Universe` in each `Type`
* store `*syntheticCache` in each `Type`
* call `t.Pointer()` which would:
  - lookup the pointer type in the Universe (careful to not create it), if found return it
  - otherwise add it to the synthetic cache (under the mutex)
  - return it

That adds 2 pointers to every Type, which is ugly but not intolerable, I think?  Every call should get the same result.

Additionally, we could add a builtin value (like `types.String`) for `Error`.

Or we could just document the "bad" behavior, I guess

yongruilin:
This SGTM. Since Universe is a map, should it be ok if we just put a non-pointer Universe in `Type`? So that we can just modify the [WalkType](https://github.com/kubernetes/gengo/blob/85fd79dbfd9fc7a328697266cbb852e1857ec2a0/v2/parser/parse.go#L639) function with something like 
```
// walkType adds the type, and any necessary child types.
func (p *Parser) walkType(u types.Universe, useName *types.Name, in gotypes.Type) (out *types.Type) {
	defer func() {
		out.Universe = u
	}()
...
```

ziyue-101:
The current encapsulation chain  is `Context -> Universe -> Packages -> Types`.

If we store `*Universe` in `Types`, would it create a circular dependency?

ziyue-101:
[...]

PR #309, opened 2025-06-25
add cache and reverse pointer to Universe to Type

ziyue-101:
Fixes:  #298


ziyue-101:
@jpbetz
 I added some unit test, Let me know if they look ok!

v2/parser/parse.go
```diff
@@ -637,6 +638,13 @@ func (p *Parser) convertSignature(u types.Universe, t *gotypes.Signature) *types
 
 // walkType adds the type, and any necessary child types.
 func (p *Parser) walkType(u types.Universe, useName *types.Name, in gotypes.Type) *types.Type {
+	var out *types.Type
+	defer func() {
+		out.Universe = &u
+		out.PointerTypeCache = &types.PointerTypeCache{
```
thockin, line 644: In my mental model you would need to pass both a "real" and "synthetic" universe around.  That is kind of ugly.  Perhaps it is simpler in the middle - a per-type object which does what I described in `multiverse` but simpler.  Still wants a mutex but doesn't need a whole synthetic `Universe`, just a single `*Package` whose name is "".  All the synthetic types have package "" (*Foo, []Foo, map[Foo]Bar)

v2/types/types.go
```diff
@@ -364,6 +372,12 @@ type Type struct {
 
 	// The underlying Go type.
 	GoType gotypes.Type
+
+	// Revevrse reference to Universe
```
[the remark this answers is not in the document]
ziyue-101: good catch, done!

v2/parser/parse.go
ziyue-101: I changed the `synthetic` to be a `map[string]*Type`, PTAL, thanks!

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+// can't be part of the main Universe because that is a simple map, with no locking, and we all know what happens
+// when you modify a map while it is being iterated. Sharing this multiverse across Types ensures that there's
+// a single *Type for every type, and maintains the invariant that PointerTo(String) == PointerTo(String).
+type multiverse struct {
+	real      Universe
+	mu        sync.Mutex
+	synthetic map[string]*Type
+}
+
+// NeMultiverse creates a multiverse.
```
yongruilin, line 384:
```suggestion
// NewMultiverse creates a multiverse.
```

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+	return t.multiverse
+}
+
+// GetOrAddType searches a Type in the Universe and synthetic map.
+// If there is a matching name, return the Type, otherwise, create the Type.
+func (m *multiverse) GetOrAddType(t *Type) *Type {
+	if m == nil {
+		return t
+	}
+	if p, ok := m.real[t.Name.Package]; ok {
```
yongruilin, line 410: This makes me wondering whether we should add the package name to the ptrType in `func PointerTo(t)`?

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+// when you modify a map while it is being iterated. Sharing this multiverse across Types ensures that there's
+// a single *Type for every type, and maintains the invariant that PointerTo(String) == PointerTo(String).
+type multiverse struct {
+	real      Universe
+	mu        sync.Mutex
+	synthetic map[string]*Type
+}
+
+// NeMultiverse creates a multiverse.
+func (t *Type) NewMultiverse(u Universe) {
```
thockin, line 385: I don't love making this a public method, since `Type` is used all over.

Why not do something like:

* Add an `Init(u)` method which panics if it is called twice.
* PointerTo() checks if the Type was initialized and panics if not.

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+	}
+	t.multiverse = &multiverse{
+		real:      u,
+		mu:        sync.Mutex{},
+		synthetic: map[string]*Type{},
+	}
+}
+
+// GetMultiverse returns a Type.multiverse.
+func (t *Type) GetMultiverse() *multiverse {
```
thockin, line 397: Should not need this if we do as above?

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+		return nil
+	}
+	return t.multiverse
+}
+
+// GetOrAddType searches a Type in the Universe and synthetic map.
+// If there is a matching name, return the Type, otherwise, create the Type.
+func (m *multiverse) GetOrAddType(t *Type) *Type {
+	if m == nil {
+		return t
```
thockin, line 408: This is wrong - it's better to let this panic with NPE than to do this

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+	t.multiverse = &multiverse{
+		real:      u,
+		mu:        sync.Mutex{},
+		synthetic: map[string]*Type{},
+	}
+}
+
+// GetMultiverse returns a Type.multiverse.
+func (t *Type) GetMultiverse() *multiverse {
+	if t == nil {
```
thockin, line 398: why not just let it panic?

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
+// a single *Type for every type, and maintains the invariant that PointerTo(String) == PointerTo(String).
+type multiverse struct {
+	real      Universe
+	mu        sync.Mutex
+	synthetic map[string]*Type
+}
+
+// NeMultiverse creates a multiverse.
+func (t *Type) NewMultiverse(u Universe) {
+	if t == nil {
```
thockin, line 386: why not just let it panic - calling methods on nil is generally not allowed

v2/types/types.go
thockin: gengo models pointers as being in package `""` - I think it's kind of weird but that is how it has always been.

v2/types/types.go
```diff
@@ -364,6 +365,60 @@ type Type struct {
...
 	GoType gotypes.Type
+
+	// The reference to Multiverse
+	multiverse *multiverse
+}
+
+// multiverse holds Type definitions that were not found in the imported code but are needed during
+// generation. For example the imported code my have T but not *T, while the generated code needs *T. This
+// can't be part of the main Universe because that is a simple map, with no locking, and we all know what happens
+// when you modify a map while it is being iterated. Sharing this multiverse across Types ensures that there's
```
thockin, line 376: The part about "Sharing this multiverse across Types..." is no longer true.

Maybe: "Storing this alongside a Type ensures that there's at most ONE *T for every T, and..."

v2/types/types.go
ziyue-101: Changed the function to be InitMultiverse(), this comment no longer applies.

v2/types/types.go
ziyue-101: * Added a panic check in InitMultiverse function.
* PointerTo() will panic if Multiverse is not initialized.

v2/types/types.go
ziyue-101: done

v2/types/types.go
ziyue-101: removed the line, now the code panics on Nil pointer

v2/types/types.go
ziyue-101: reworded, thanks!

v2/types/types.go
```diff
@@ -364,6 +365,50 @@ type Type struct {
...
+
+	// The reference to Multiverse
+	multiverse *multiverse
+}
+
+// multiverse holds Type definitions that were not found in the imported code but are needed during
+// generation. For example the imported code my have T but not *T, while the generated code needs *T. This
+// can't be part of the main Universe because that is a simple map, with no locking, and we all know what happens
+// when you modify a map while it is being iterated. Storing the multiverse alongside a Type ensures that there's
+// at most one *Type for every type, and maintains the invariant that PointerTo(String) == PointerTo(String).
```
thockin, line 377: Super not since I am leaving other comments.  s/String/T in the invariant

v2/types/types.go
```diff
@@ -364,6 +365,50 @@ type Type struct {
...
+type multiverse struct {
+	real      Universe
+	mu        sync.Mutex
+	synthetic map[string]*Type
+}
+
+// InitMultiverse inits a multiverse for a Type.
+// It panics if called twice.
+func (t *Type) InitMultiverse(u Universe) {
```
thockin, line 386: Just call it Init -- multiverse is an implementation detail, it doesn't need to leak into the API

v2/types/types.go
```diff
@@ -364,6 +365,50 @@ type Type struct {
...
+	if t.multiverse != nil {
+		panic("Can't initialize a non-empty multiverse on Type")
```
thockin, line 388: "can't re-initialize type %v"...

v2/types/types_test.go
```diff
@@ -37,6 +39,95 @@ func TestGetBuiltin(t *testing.T) {
...
+		},
+	}
+
+	type3 := &Type{
+		Name: Name{Package: "pkgname", Name: "thridstructname"},
+		Kind: Struct,
+		multiverse: &multiverse{
+			real: u,
+			synthetic: map[string]*Type{
+				"*pkgname.thridstructname": &Type{
```
thockin, line 75: s/thrid/third

Or just call these T1, T2, T3 :)

v2/parser/parse.go
```diff
@@ -637,6 +637,10 @@ func (p *Parser) convertSignature(u types.Universe, t *gotypes.Signature) *types
 
 // walkType adds the type, and any necessary child types.
 func (p *Parser) walkType(u types.Universe, useName *types.Name, in gotypes.Type) *types.Type {
+	var out *types.Type
+	defer func() {
+		out.InitMultiverse(u)
```
thockin, line 642: I just realized someone needs to call this on all of the builtin types.  This means that there can be only 1 Universe because the builtins are global variables, which have a single multiverse pointer each.  That makes me sad.

Can you think of any alternative that is not so limited?

Is the cure worse than the sickness?  I don't like that PointerTo(T) != PointerTo(T) but if there were two universes active T != T, anyway...

Have to think on this...

thockin: Revisiting this - I still don't know what we should do.

Large-scale redesign of gengo apps (e.g. making the Universe the central abstraction, passing it everywhere, so instead of `types.String` you might use `u.String()` or `u.PointerTo(t)`) seems unlikely.

Making a global universe (so there can't be more than one) seems plausible but amounts to gengo v3, and I don't see anyone working on that right now, especially without some other driver.

We could pre-populate a `t.Pointer` up to some level of depth (e.g. ***T) and thast "should be enough for anyone" but requires more gengo work.

I'm letting this sit for now until it is clearer what we need.

PR #311, opened 2025-07-06
Add an option to forbid tag arguments

gavinkflam:
Part of kubernetes/kubernetes#130358

`ExtractCommentTags` is deprecated. Old tags are not in function style so the migration to `ExtractFunctionStyleCommentTags` is not straightforward. This new option offers an easier migration path to parse non-function style tags.

We have agreed that we want to be strict and have the code generator fail with an error if an unexpected argument is present. Refer to the discussion [here](https://github.com/kubernetes/kubernetes/pull/131711/files#r2132934997).

This can unblock the migration of `ExtractCommentTags` usages in kube-openapi, and eliminate an utility function introduced in kubernetes/kubernetes#131711.

cc @jpbetz @yongruilin

### DIFF
--- a/v2/comments.go
+++ b/v2/comments.go
@@ -116,7 +116,7 @@ func ExtractFunctionStyleCommentTags(marker string, tagNames []string, lines []s
 			if err != nil {
 				return nil, err
 			}
-			tag, err := toStringArgs(typedTag)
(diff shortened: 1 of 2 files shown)


gavinkflam:
cc @jpbetz @yongruilin

/remove-lifecycle stale

PR #315, opened 2025-09-22
Fix gofmt missing case in code generator

Jefftree:
Fix gofmt missing case in code generator. Fixes https://github.com/kubernetes/kubernetes/issues/134207

/assign @thockin @BenTheElder


thockin (commented):
Damn, thanks for that.

/lgtm
/approve

PR #316, opened 2025-11-10
Bump dependencies to be compatible with Go 1.25

ash2k:
I'm getting the following error with the current versions. Works fine with the updated versions. Kubernetes is likely not affected because it probably uses newer versions of these libraries anyway (I haven't checked).

```
/home/ci/go/pkg/mod/golang.org/x/tools@v0.16.1/internal/tokeninternal/tokeninternal.go:78:9: invalid array length -delta * delta (constant -256 of type int64)
```


v2/go.mod
```diff
@@ -1,15 +1,16 @@
 module k8s.io/gengo/v2
 
-go 1.23
+go 1.24.0
```
ash2k, line 3: Required by the new version of `golang.org/x/tools`.

ash2k:
PTAL @thockin @sttts

thockin:
Thanks!

/lgtm
/approve

PR #317, opened 2026-04-08
fix(tags): detect inline fields structurally instead of matching ",inline" tag

aramase:
Determine whether a field should be inlined based on whether it is embedded and has an empty json tag name, rather than checking for the ",inline" option which is not a real encoding/json directive.

This allows types to use `json:""` instead of `json:",inline"` on embedded fields like TypeMeta, ObjectMeta, and ListMeta.

xref: https://github.com/kubernetes/kubernetes/pull/138260


jpbetz (approved):
/approve
/lgtm

Looking forward to getting this cleaned up.

